PR #15, opened 2023-11-04
Updated Confetti sample in multiplatform-samples.md

arkivanov:
Confetti sample now uses Decompose. Updated multiplatform-samples.md, replaced `kmm-viewmodel` with `decompose`.

cc: @joreilly


koshachy (approved):
Thanks for pointing out!

PR #18, opened 2023-11-07
Update compose-compatibility-and-versioning.md

eymar:
The table was missing a Compose Mutliplatform compiler plugin version for Kotlin 1.9.20


danil-pavlov (approved):
Thank you!

PR #20, opened 2023-11-08
Change name of MainActivity extension from .k to .kt

eemjema:
When I was going throught the documentation I realized that a ´t´ is missing from MainActivity.k. This is a minor issues but I thought it might help other beginners in the future.


danil-pavlov (approved):
Definitely, thank you for noticing and for your report!

PR #35, opened 2023-12-01
New and improved sourceSets declaration

zamulla:
As advised in https://youtrack.jetbrains.com/issue/KT-60734


danil-pavlov (approved):
Thanks!
The common source set just needs the same change on the previous step, Add dependencies

PR #53, opened 2024-01-11
Update the style sample code

Hailong:
The padding for Column is missing.


zamulla (approved):
Thanks for noticing that!

PR #57, opened 2024-01-22
Fix sample code

lindenquan:


SebastianAigner (approved):

SebastianAigner:
Yup, that fixes the problem. Thanks for noticing and contributing!

zamulla (approved):
Thanks for noticing that!

PR #58, opened 2024-01-24
KMPizza added to the Samples page

zamulla:


koshachy (approved):
Approved.
Still hesitating about the indentations of HTML tags but it is not important, I guess 😄

PR #60, opened 2024-01-26
update: added appropriate Fleet tutorial links

zamulla:
In the Compose tutorial as well as in a FAQ about IDEs.

### DIFF
--- a/topics/compose-onboard/compose-multiplatform-getting-started.md
+++ b/topics/compose-onboard/compose-multiplatform-getting-started.md
@@ -13,6 +13,8 @@ events, and modifiers.
   this tutorial.
 * To complete this tutorial, you'll only need Android Studio. It allows you to try multiplatform development on Android
   and desktop. For iOS, you need a macOS machine with XCode installed. This is a general limitation of iOS development.
+  * To try a new multiplatform experience without juggling different IDEs and switching to
+  Xcode for writing Swift code, try out [JetBrains Fleet for multiplatform development](fleet.md).
 * If you wish, you can limit your choice to the specific platforms you're interested in and omit the others.
 
 Get to know Compose Multiplatform by completing these steps:
--- a/topics/compose-onboard/compose-multiplatform-setup.md
+++ b/topics/compose-onboard/compose-multiplatform-setup.md
@@ -13,6 +13,9 @@
 Before you create your first application that works on both iOS and Android, you'll need to set up an environment for Kotlin Multiplatform
 development.
 
+To try a new multiplatform experience without juggling different IDEs and switching to
+Xcode for writing Swift code, try out JetBrains Fleet following [a similar tutorial](fleet.md).
+
 ## Install the necessary tools
 
 We recommend that you install the latest stable versions for compatibility and better performance.
--- a/topics/compose-onboard/compose-multiplatform-create-first-app.md
+++ b/topics/compose-onboard/compose-multiplatform-create-first-app.md
@@ -28,6 +28,8 @@ To begin, create a sample project. This is best achieved with the Kotlin Multipl
 ## Examine the project structure
 
 1. Launch Android Studio.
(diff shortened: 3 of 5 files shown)


topics/compose-onboard/compose-multiplatform-setup.md
```diff
@@ -13,6 +13,9 @@
 Before you create your first application that works on both iOS and Android, you'll need to set up an environment for Kotlin Multiplatform
 development.
 
+To try a new multiplatform experience without juggling different IDEs and switching to
```
danil-pavlov, line 16: Id' make at a tip, just like it Get started. It helps to separate from the rest of the text, making it a side recommendation

topics/compose-onboard/compose-multiplatform-setup.md
```diff
@@ -142,6 +145,13 @@ To make sure everything works as expected, install and run the KDoctor tool:
               <li><code>command not found: brew</code> — <a href="https://brew.sh/">install Homebrew</a>.</li>
               <li><code>command not found: java</code> — <a href="https://www.oracle.com/java/technologies/javase-downloads.html">install Java</a>.</li>
            </list>
+   </def>
+   <def title="Subpar multiplatform development experience">
```
danil-pavlov, line 149: Not sure if this fits into Possible issues and solutions, maybe we should just leave the Fleet note above. @koshachy WDYT?

Also, "command not found" issues seem irrelevant for this part, or am I missing something?

koshachy: Well, it could be a possible issue so we can leave it here.
I'd recommend changing the title to something more simpler than now and make it more useful to the end user.
We could brainstorm here, but at the first glance we could make it in this way:

**Still having trouble setting up your environment?**

Create a YouTrack issue ...

Also, you could try new JetBrains Fleet ...

or

**Is your problem not on the list?**

Create a YouTrack issue ...

Also, you could try new JetBrains Fleet ...

topics/compose-onboard/compose-multiplatform-getting-started.md
```diff
@@ -8,11 +8,16 @@ In this tutorial, you will build a sample application that runs on Android, iOS,
 interface, you will use the Compose Multiplatform framework and learn about its basics: composable functions, themes, layouts,
 events, and modifiers.
 
-* No previous experience with Compose Multiplatform, Android, or iOS is required for this tutorial. We recommend that
-  you become familiar with the [fundamentals of Kotlin](https://kotlinlang.org/docs/getting-started.html) before starting
-  this tutorial.
+> To try a new multiplatform experience without juggling different IDEs and switching to
+> Xcode for writing Swift code, try out [a similar tutorial for JetBrains Fleet](fleet.md).
```
koshachy, line 12: Some concerns here:
* Well, are they actually similar? It seems that they are not, so I'd recommend removing the accent of this tip, from "similar tutorial" to "new multiplatform experience" only.
* It seems more logical for me, if this tip will be located below "Things to keep in mind for this tutorial" list. WDYT?

topics/compose-onboard/compose-multiplatform-setup.md
zamulla: I tried making the entry more useful here: https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/60/commits/29e354dce1311d327ea815125cc6b7510fc04e7d

topics/compose-onboard/compose-multiplatform-getting-started.md
zamulla: > are they actually similar
You're right, rephrased this.

> below "Things to keep in mind for this tutorial" list
We considered that, and wanted to mention Fleet before "you'll only need Android Studio". But the list is universal otherwise, and putting the tip below does make more sense.

koshachy (approved):
LGTM!

PR #59, opened 2024-01-26
Add MusicApp-KMP to samples

SEAbdulbasit:
I wanted to add Music App (https://github.com/SEAbdulbasit/MusicApp-KMP) to the samples list. This app demonstrates how to use native APIs in KMP. It retrieves the top 50-chart list from Spotify API and utilizes native APIs to load music from Spotify APIs. Ktor is used for network calling and Decompose for navigation. It supports Android, iOS, Web, and Desktop platforms.

PR #68, opened 2024-02-07
Add TravelApp-KMP to KMP samples

SEAbdulbasit:
Hi, 
Wanted to add TravelApp-KMP to the samples. You can find that live here: https://seabdulbasit.github.io/TravelApp-KMP/
It is a simple UI showcase project.

Here is the repo link: https://github.com/SEAbdulbasit/TravelApp-KMP

### DIFF
--- a/topics/multiplatform-samples.md
+++ b/topics/multiplatform-samples.md
@@ -1,12 +1,15 @@
 [//]: # (title: Kotlin Multiplatform samples)
 
 This is a curated list of cross-platform projects created with Kotlin Multiplatform. Some share almost all their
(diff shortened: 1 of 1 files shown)


SebastianAigner:
Hi @SEAbdulbasit, thanks for the PR. We try to keep the number of examples for particular areas somewhat limited. I believe your music app with the points mentioned in #59 would be a great fit, but I think the Travel App is a bit too barebones for our sample gallery, and another app of yours, the Recipe App (that is already part of this list) has a lot of the same areas already covered very well. Thanks again for contributing that one! 🎉 

So I hope you'll understand that I'll close this PR for now :)

PR #72, opened 2024-02-14
Add kotlin-multiplatform-sample topic

SebastianAigner:
The `kotlin-multiplatform` topic is dominated by much-used libraries, which makes it harder to discover sample projects even if they have a couple hundred starts. This PR adds a link to the "kotlin-multiplatform-sample" topic (which currently isn't populated), which can be used for this specific use case.


zamulla (approved):
Code is green.

PR #78, opened 2024-02-20
fix: change the syntax of the lifecycle-viewmodel-compose dependency

zamulla:
Different styles of dependencies don't work well with copy-paste. This is a quick fix for the tutorial experience. In the future, we'll probably update this step to use the version catalog explicitly.


koshachy (approved):
LGTM!

PR #89, opened 2024-03-13
update: separate iOS integration into articles for UIKit and SwiftUI

zamulla:


topics/compose/compose-swiftui-integration.md
```diff
@@ -0,0 +1,118 @@
+[//]: # (title: Integration with the SwiftUI framework)
+
+Compose Multiplatform is interoperable with the [SwiftUI](https://developer.apple.com/xcode/swiftui/) framework.
+You can embed Compose Multiplatform within a SwiftUI application as well as embed native SwiftUI components within
+Compose Multiplatform UI. This page provides examples of how you can do this:
+
+* [Use Compose Multiplatform inside a SwiftUI application](#use-compose-multiplatform-inside-a-swiftui-application)
```
danil-pavlov, line 7: I don't think we need these lists anymore on both pages. Articles are relatively short now, plus we have the right-hand menu with the same elements.

danil-pavlov (approved):

PR #93, opened 2024-03-21
update: Add Collective Idea / Farmish to case studies

darronschall:
Collective Idea is custom software consultancy well versed in KMP that recently delivered the Farmish iOS and Android apps.


zamulla:
Hi! Thank you for reaching out, and sorry for a late reply.

Changes to the list of case studies are almost always initiated by JetBrains, to keep it both manageable and representative. We will keep your example in mind for the future, though.

PR #108, opened 2024-04-19
feat: move Multiplatform journal

koshachy:
[KT-66986](https://youtrack.jetbrains.com/issue/KT-66986)


images/journal/survey-results-q1-q2-21.png
zamulla, line 1: There's a mention of KMM here, and the risky approach of a color gradient in a pie chart. Not a huge deal, but maybe we can fix it in a separate issue.

zamulla (approved):
👍

PR #109, opened 2024-04-21
Fix callback names in lifecycle table

zsmb13:
Two entries in the table were swapped. Gaining focus resumes the Desktop app, while losing it makes it paused.


zamulla (commented):
👍

zamulla (approved):

PR #116, opened 2024-04-30
update Lifecycle events mapping on iOS after changes in 1.6.0-beta3

kropp:
See https://github.com/JetBrains/compose-multiplatform-core/pull/1319


zamulla (approved):
👍

PR #120, opened 2024-05-05
Update User Interface for KMM Sample

myabc:
Kotlin Multiplatform Mobile Sample uses Jetpack Compose for UI as of commit: https://github.com/Kotlin/kmm-basic-sample/commit/2c22d9140bde535f9c2f8aabf5ab8f1900868c35


zamulla (approved):
Thanks for noticing!

PR #123, opened 2024-05-08
Spm local integration

danil-pavlov:


topics/development/multiplatform-spm-local-integration.md
```diff
@@ -0,0 +1,128 @@
+[//]: # (title: Using Kotlin from local Swift packages)
+
+In this tutorial, you'll learn how to add a Kotlin module to a local package in the Swift package manager.
+This is useful if you want to simultaneously connect your Kotlin Multiplatform module to several native iOS modules
+in your Xcode project.
+
+## Set up the project
```
daugeldauge, line 7: Any version newer than `2.0.0-Beta2` can be used now. So I think this section can be replaced with a small note to use the latest 2.0.0 release candidate.

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -0,0 +1,128 @@
+[//]: # (title: Using Kotlin from local Swift packages)
+
+In this tutorial, you'll learn how to add a Kotlin module to a local package in the Swift package manager.
+This is useful if you want to simultaneously connect your Kotlin Multiplatform module to several native iOS modules
```
daugeldauge, line 4: That's slightly different — the new thing is that we're now able to depend on the Kotlin framework in SPM packages. Previously, it was not possible to do this even in one module.

daugeldauge, line 4: > native iOS module

Not sure if it's a well-understood terminology. At least for me, it's a bit confusing.

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -0,0 +1,128 @@
...
+>
+{type="tip"}
+
+## Remove previous integration
+
+If you're migrating your project from another integration approach, remove it first:
+
+#### CocoaPods plugin {initial-collapse-state="collapsed"}
+
+> If you have dependencies on other Pods in the `cocoapods` block, you have to keep to the CocoaPods integration approach.
```
daugeldauge, line 39: Maybe it's worth expanding here, that it means that the tutorial has no use in this case (meaning that it's impossible for now to depend on pods and reference Kotlin in a multi-module SPM project at the same time)

daugeldauge (changes requested):
We also may have to add an additional step with `embedAndSign` invocation in the build phase to fix the build for a real device. I'll add the details a bit later.

danil-pavlov:
@daugeldauge, thank you for all the suggestions! Please check the new version

daugeldauge (approved):

zamulla (approved):
👍

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -0,0 +1,99 @@
...
+
+   ```bash
+   cd "$SRCROOT/.."
+   ./gradlew :shared:embedAndSignAppleFrameworkForXcode
+   ```
+4. Choose your app's target in the **Provide build settings from** section:
+
+   ![Filled run script action](xcode-filled-run-script-action.png){width=700}
+
+5. Use the code from Kotlin in the local Swift package added to your Xcode project:
```
markst, line 85: Could be worth mentioning about explicitly linking the dependency in the Local Swift Package?
Not sure if project will build if “Find Implicit Dependencies” is disable. 

FYI: https://github.com/swiftlang/swift-package-manager/issues/5297

markst: @daugeldauge so you mean a SPM plugin would be required to prebuild the xcframework in order to be linked? Without using the Run Script Phase.

Using these steps if we were to open the `Package.swift` independently, the package will not compile. 
I believe there needs to be some explanation of this in the readme to explain Xcode finds Implicit Dependencies.

---

Perhaps it's a alternative approach, but manual usage of the assembled framework might be as follows:

5. Build the `XCFramework` using:
   ```bash
   ./gradlew assembleXCFramework
   ```
   This will generate `MyLibrary.xcframework` in the `build/XCFrameworks/release` and `build/XCFrameworks/debug` directories.

6. Link the generated `MyLibrary.xcframework` in your `Package.swift` file:

   ```swift
   // swift-tools-version:5.3
   import PackageDescription

   let package = Package(
       name: "MyPackage",
       platforms: [
           .iOS(.v13)
       ],
       products: [
           .library(
               name: "MyLibrary",
               targets: ["MyLibrary"]
           ),
       ],
       dependencies: [
           // Add the path to the xcframework
           .binaryTarget(
               name: "MyLibrary",
               path: "../path/to/MyLibrary.xcframework"
           ),
       ],
   )
   ```

markst: Something like this?:https://github.com/marty-suzuki/kmm-with-spm-buildtoolplugin-sample/blob/main/iosApp/Modules/plugins/MultiplatformBuildPlugin/MultiplatformBuildPlugin.swift

I believe the sandboxing is only an issue when attempting to write to arbitrary locations in the file system, where as if we perform as task on gradle, I'm not sure it matters? I'll need to investigate

PR #127, opened 2024-05-12
Fix enabling Developer Mode instructions for iOS

myabc:


zamulla (approved):
Thanks for catching that!

PR #129, opened 2024-05-13
fix: three repos illustration

danil-pavlov:


images/config-options/three-repositories.svg
zamulla, line 1: The lines going inside the "Android project" and the "Xcode project" bubbles look like they end in the wrong place, intuitively they should connect with similar lines around the project blocks.

To keep the visual consistent, maybe switch around the styles? So that the "Android app", for example, has the solid border, and the "Android project" is the gray block without the border.

zamulla (changes requested):

images/config-options/three-repositories.svg
danil-pavlov: That was all the designer's decision with the lines and labels, and it's consistent with other diagrams on this page. I can also argue that lines should connect apps (frameworks) and not projects, as they do in case of mono- and two repos. "Android app" and the "Android project" are also different concepts, and it's important to show the difference here when you have a separate Android project

zamulla: I finally got to finding the page, all of them are like this, I see :)

zamulla (approved):
👍

PR #133, opened 2024-05-17
update: rename get started in tutorials

ElviraMustafina:
a temporary workaround for the duplicated Get Started issue: it's bad timing for a revolution with removing pages and changing links, but we can try renaming instead


Daria-Voronina (approved):
Sorry for the delayed reply! Thank you for the proposed changes. Looks good.

PR #134, opened 2024-05-20
update: update Fleet documentation

koshachy:


topics/tools/fleet.md
```diff
@@ -83,9 +84,9 @@ In this tutorial, you'll create a new project with the web wizard:
 
    At the top of the window, you'll see messages from Fleet as the project is imported and indexed.
 
-When Smart Mode is enabled, Fleet operates as an IDE, with all of the code completion, navigation, debugging, and
-refactoring features available. When it's disabled, Fleet works as a simple file editor. You can quickly open files and
-make changes, but most of the productivity features won't be available.
+When Smart Mode is enabled, Fleet operates as a _polyglot code editor_, that offers language-related features such as code completion,
```
zamulla, line 87:
```suggestion
When Smart Mode is enabled, Fleet offers language-related features such as code completion,
```

As I see it, the difference between a polyglot code editor and a simple code editor is not as self-evident as it needs to be to illustrate the power of Smart Mode.

Also, do you think we should talk about the enabled Smart Mode in the end as opposed to the beginning? RIght now, it reads "smart mode is good, but without it it's also fine", when it might have been "without smart mode you're fast, and with smart mode you're powerful".

zamulla (approved):
Looks good, one comment to consider.

topics/tools/fleet.md
koshachy: Yeah, the second option is what we want.
I think we can back to it a bit after, to rephrase the whole section.

PR #138, opened 2024-05-21
Update compose-navigation-routing.md

egorikftp:


zamulla:
Hi! Thank you for reaching out, and sorry for the late reply.

The list of Compose navigation solutions is meant for more widely adopted libraries, especially now that Compose Multiplatform implements baseline Navigation from Jetpack Compose. We will keep an eye on you though!

PR #144, opened 2024-05-23
update: separate page for the common ViewModel

zamulla:


topics/compose/compose-viewmodel.md
```diff
@@ -0,0 +1,63 @@
...
+    @Composable
+    fun CupcakeApp(
+        viewModel: OrderViewModel = viewModel { OrderViewModel() },
+    ) {
+    // ...
+    }
+    ```
+   
+On non-JDK platforms objects cannot be instantiated using type reflection.
+So in common code you cannot call the `viewModel()` function without parameters: every time a ViewModel is created,
```
MatkovIvan, line 58:
```suggestion
So in common code you cannot call the `viewModel()` function without parameters: every time a `ViewModel` is created,
```

topics/compose/compose-viewmodel.md
```diff
@@ -0,0 +1,63 @@
...
+you need to provide at least an initializer as an argument.
+
+If only an initializer is provided, Compose Multiplatform creates a default factory under the hood.
```
MatkovIvan, line 61: > Compose Multiplatform creates

it's not correct. It does not relate to Compose

MatkovIvan (changes requested):

topics/compose/compose-viewmodel.md
```diff
@@ -0,0 +1,63 @@
...
+    ```
+
+2. Add the ViewModel to your composable function:
+
+    ```kotlin
+    @Composable
+    fun CupcakeApp(
+        viewModel: OrderViewModel = viewModel { OrderViewModel() },
+    ) {
+    // ...
```
MatkovIvan, line 53:
```suggestion
        // ...
```

topics/compose/compose-viewmodel.md
```diff
@@ -0,0 +1,63 @@
...
+    }
+    ```
+   
+On non-JDK platforms objects cannot be instantiated using type reflection.
+So in common code you cannot call the `viewModel()` function without parameters: every time a `ViewModel` is created,
+you need to provide at least an initializer as an argument.
+
+If only an initializer is provided, the library creates a default factory under the hood.
+But you can implement your own factories and call more explicit versions of the common `viewModel()` function,
```
MatkovIvan, line 62:
```suggestion
But you can implement your own factories and call more explicit versions of the common `viewModel` function,
```

topics/compose/compose-viewmodel.md
```diff
@@ -0,0 +1,63 @@
...
+To use the multiplatform `ViewModel` implementation, add the following dependency to your `commonMain` source set:
+
+```kotlin
+kotlin {
+  // ...
+  sourceSets {
+    // ...
+    commonMain.dependencies {
+      // ...
+      implementation("org.jetbrains.androidx.lifecycle:lifecycle-viewmodel-compose:%composeViewmodelVersion%")
```
MatkovIvan, line 21: `%composeViewmodelVersion%` should be updated to `2.8.0`

MatkovIvan (approved):

koshachy (approved):
Approved.
Resolve the comments before merging.

PR #150, opened 2024-05-29
chore: a link to the SPM export setup page

danil-pavlov:


koshachy (approved):
LGTM.
BTW, I think we could use short versions of the YouTrack issues. I see that it's not your change.

PR #160, opened 2024-06-07
Add Rijksmuseum to multiplatform-samples

fethij:
Hallo, Rijksmuseum is Kotlin and Compose Multiplatform app that offers an immersive way to explore the art collection of the renowned Rijksmuseum in Amsterdam. The project is multi-modular and uses compose for all platforms (Android, iOS, desktop and web). It also showcases topics like DI, Networking, Navigation, Caching and Sharing build logic between subprojects/modules...

If this is a good sample for people, please include it in the doc, or else feel free to close the PR.

Here is the repo: https://github.com/fethij/rijksmuseum

### DIFF
--- a/topics/multiplatform-samples.md
+++ b/topics/multiplatform-samples.md
@@ -523,4 +523,36 @@ to add your Kotlin Multiplatform project to these topics and help the community.
           </list>
       </td>
   </tr>
+  
+  <tr>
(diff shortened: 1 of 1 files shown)


zamulla (approved):
Sorry for the delay, adding it now 👌

PR #170, opened 2024-06-25
Fix incorrect resource `readBytes` name

alexzhirkevich:


terrakok (approved):

zamulla (approved):
Thanks for catching it!

PR #177, opened 2024-07-14
fix string array usage wrong example

wiryadev:


zamulla:
Thank you for pointing it out! We fixed it somewhere else, I think, before I got to the PR.

PR #179, opened 2024-07-15
Update multiplatform-upgrade-app.md

DevAlloy:
I'm not sure how to properly identify the 0.8.2 version, but we need a new version with K2 support


zamulla:
Thanks for pointing it out! I'll update the SKIE version in the docs (variables are stored in the `v.list` file at the root of the project).

The `async` part seems to be required only with Xcode 16, we'll see if that remains a breaking change when this version of the IDE is out of beta.

PR #183, opened 2024-07-16
fix: clarifying the migration steps

zamulla:
Here we're trying to make sure that users don't take unnecessary steps during migration: referencing Compose compiler Maven artifacts directly is only necessary if there is no Gradle plugin to resolve them.

### DIFF
--- a/topics/compose/compose-compiler.md
+++ b/topics/compose/compose-compiler.md
@@ -72,7 +72,8 @@ To resolve this, delete all of the `build` directories: at the root of your proj
 ## Migrating a Jetpack Compose project
 
 When migrating to Kotlin 2.0.0 or newer from 1.9, you should adjust your project configuration depending on the way you deal with
-the Compose compiler now. We recommend using the Kotlin Gradle plugin and the Compose compiler Gradle plugin for best results. 
+the Compose compiler now. We recommend using the Kotlin Gradle plugin and the Compose compiler Gradle plugin
+to automate configuration management. 
 
 ### Managing the Compose compiler with Gradle plugins
 
--- a/topics/compose/compose-compiler.md
+++ b/topics/compose/compose-compiler.md
@@ -107,8 +107,8 @@ For Android modules that do not rely on Compose Multiplatform:
 4. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
    See [Compiler options](#compose-compiler-options-dsl) for reference.
 
-> If you are not using Gradle plugins to manage the Compose compiler, you should update the direct references to old Maven
-> artifacts:
+> If you are not using Gradle plugins to manage the Compose compiler, update any direct references to old Maven
+> artifacts in your project:
 > 
 >   * Change `androidx.compose.compiler:compiler` to `org.jetbrains.kotlin:kotlin-compose-compiler-plugin-embeddable`
 >   * Change `androidx.compose.compiler:compiler-hosted` to `org.jetbrains.kotlin:kotlin-compose-compiler-plugin`
--- a/topics/compose/compose-compiler.md
+++ b/topics/compose/compose-compiler.md
@@ -71,6 +71,9 @@ To resolve this, delete all of the `build` directories: at the root of your proj
 
 ## Migrating a Jetpack Compose project
 
+When migrating to Kotlin 2.0.0 or newer from 1.9, you should adjust your project configuration depending on the way you deal with
+the Compose compiler now.
+
 ### Managing the Compose compiler with Gradle plugins
 
 For Android modules that do not rely on Compose Multiplatform: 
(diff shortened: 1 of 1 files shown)


topics/compose/compose-compiler.md
```diff
@@ -109,18 +104,16 @@ For Android modules that do not rely on Compose Multiplatform:
...
-4. If you reference the old Maven artifacts directly, you'll need to update these references:
-
-   * Change `androidx.compose.compiler:compiler` to `org.jetbrains.kotlin:compose-compiler-gradle-plugin-embeddable`
-   * Change `androidx.compose.compiler:compiler-hosted` to `org.jetbrains.kotlin:compose-compiler-gradle-plugin`
-
-5. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
+4. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
    See [Compiler options](#compose-compiler-options-dsl) for reference.
 
-   > The Gradle plugin provides defaults for several compiler options that were only specified manually before.
```
ShikaSD, line 120: I think this block should stay, since it is an important piece of information.

ShikaSD: Nvm, noticed it below.

zamulla: Yes, I realized there's no need to copy-paste it here. People still need to see the reference if the step is relevant to them, so let's put it in the reference.

topics/compose/compose-compiler.md
```diff
@@ -109,18 +104,16 @@ For Android modules that do not rely on Compose Multiplatform:
...
-   > If you have any of them set up with `freeCompilerArgs`, for example, Gradle will report a duplicate options error.
-   >
-   {type="warning"}
+> If you are not using Gradle and thus unable to enjoy Gradle plugins, you should update the direct references to old Maven
```
ShikaSD, line 110: Technically, some people are using Gradle, but are also configuring the compiler on their own :)

```suggestion
> If you are configuring Compose compiler plugin on your own, you should update the references to Maven artifacts:
```

zamulla: Just to make sure: are they configuring the compiler, or the compiler plugin? I corrected in a slightly different way, tell me if that is less correct: https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/183/commits/7a25c8c13aaffcee96bf5cb410ae57e18fe9a995

ShikaSD: Technically, Compose compiler plugin, but we use these interchangeably in docs.

ShikaSD (commented):
Lgtm, thanks!

Tapchicoma (approved):

topics/compose/compose-compiler.md
```diff
@@ -109,18 +104,16 @@ For Android modules that do not rely on Compose Multiplatform:
...
-5. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
+4. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
    See [Compiler options](#compose-compiler-options-dsl) for reference.
 
-   > The Gradle plugin provides defaults for several compiler options that were only specified manually before.
-   > If you have any of them set up with `freeCompilerArgs`, for example, Gradle will report a duplicate options error.
-   >
-   {type="warning"}
+> If you are not using Gradle plugins to manage the Compose compiler, you should update the direct references to old Maven
+> artifacts:
```
sarahhaggarty, line 111:
```suggestion
> If you are not using Gradle plugins to manage the Compose compiler, update any direct references to old Maven
> artifacts in your project:
```

topics/compose/compose-compiler.md
```diff
@@ -109,18 +104,16 @@ For Android modules that do not rely on Compose Multiplatform:
...
-
-5. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
+4. If you are using compiler options for the Jetpack Compose compiler, set them in the `composeCompiler {}` block.
    See [Compiler options](#compose-compiler-options-dsl) for reference.
 
-   > The Gradle plugin provides defaults for several compiler options that were only specified manually before.
-   > If you have any of them set up with `freeCompilerArgs`, for example, Gradle will report a duplicate options error.
-   >
-   {type="warning"}
+> If you are not using Gradle plugins to manage the Compose compiler, you should update the direct references to old Maven
```
sarahhaggarty, line 110: If I understand correctly, the instructions are about adding Gradle plugins for the Compose compiler, so maybe it is better to have this note at the beginning of the section? Otherwise I worry that readers will not see it.

topics/compose/compose-compiler.md
sarahhaggarty: Project or module? Not sure but it would be nice to specify the area that they should check :)

sarahhaggarty (approved):
Looks good! Just some minor suggestions for you to consider.

topics/compose/compose-compiler.md
zamulla: The problem was that too many people saw it before and applied to themselves without thinking :)

I tried to delineate the use cases with subtitles, please tell me what you think about that.

PR #192, opened 2024-07-25
update: improvements to the local spm tutorial

danil-pavlov:


topics/development/multiplatform-spm-local-integration.md
```diff
@@ -13,6 +25,11 @@ in your Xcode project.
 
 The feature is available starting with Kotlin 2.0.0.
 
+> To check the Kotlin version, navigate to the `build.gradle(.kts)` file in the root of your Kotlin Multiplatform project.
```
DevAlloy, line 28: or in gradle/libs.versions.toml

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -65,45 +81,51 @@ To migrate from the CocoaPods plugin:
...
+4. Delete the `.podspec` and `Podfile` files.
 
 ## Connect the framework to your project
 
 > The integration into `swift build` is currently not supported.
 >
 {type="note"}
 
-To connect the Kotlin framework to your Xcode project and use the Kotlin code in a local Swift package:
+To connect the iOS part of your Kotlin Multiplatform project to your Xcode project and use the Kotlin code in a local Swift package:
```
DevAlloy, line 92: Technically, the artifact will not only have iOS, but the common part as well

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -65,45 +81,51 @@ To migrate from the CocoaPods plugin:
...
 
-1. In Xcode, go to **Product** | **Scheme** | **Edit scheme** or click the schemes icon in the top bar and select **Edit scheme**:
+1. In Xcode, open your iOS project and the local SPM module you want to use for integration with the Kotlin framework.  
+2. Go to **Product** | **Scheme** | **Edit scheme** or click the schemes icon in the top bar and select **Edit scheme**:
```
DevAlloy, line 95: The previous version was fine because we need to edit the iOS project scheme.

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -13,6 +25,11 @@ in your Xcode project.
 
 The feature is available starting with Kotlin 2.0.0.
 
+> To check the Kotlin version, navigate to the `build.gradle(.kts)` file in the root of your Kotlin Multiplatform project.
+> You'll see the current version in the `plugins {}` block at the top of the file.
+> 
+{type="tip"}
+
 If you still haven't upgraded to Kotlin 2.0.0, you can use a special Kotlin version, `1.9.24-spm2` from the
```
DevAlloy, line 33: This is a custom build that we used only for UX sessions. It is better not to mention it in the final version.

topics/development/multiplatform-spm-local-integration.md
danil-pavlov: Thank you, rephrased the whole sentence

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -1,8 +1,20 @@
...
+
+This is a local integration method that can work for you if:
+
+* You have an iOS app with local SPM modules.
+* You've already set up a Kotlin Multiplatform project targeting iOS on your local machine.
+
+![Direct integration diagram](direct-integration-scheme.svg){width="700"}
+
+In contrast with a standard direct integration method used in project wizards, here you'll add the
+`embedAndSignAppleFrameworkForXcode` task not to the build phase but as a pre-action in your project's build settings.
```
zamulla, line 14: Is it really beneficial to mention the project wizard here?

I'd start with the profit ("You can set up the package so that changes in common code are reflected in Swift instantly") and then explain the mechanism ("because you'll execute the X task as a pre-action, and not as a build phase").

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -1,8 +1,20 @@
...
+
+This way, you don't need to build both the Kotlin Multiplatform and the iOS project to see the changes made in the common
```
zamulla, line 16: With the negative, it's not clear what the process is: you don't need to build both projects, but what do you need to build? Do you need to build anything?

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -65,16 +63,16 @@ To migrate from the CocoaPods plugin:
...
+4. Delete the `.podspec` and `Podfile` files.
 
 ## Connect the framework to your project
 
 > The integration into `swift build` is currently not supported.
 >
 {type="note"}
 
-To connect the Kotlin framework to your Xcode project and use the Kotlin code in a local Swift package:
+To be able to use Kotlin code in a local Swift package, now connect the Kotlin framework generated from the multiplatform
```
zamulla, line 74:
```suggestion
To be able to use Kotlin code in a local Swift package, connect the Kotlin framework generated from the multiplatform
```

images/migrate-app/direct-integration-scheme.svg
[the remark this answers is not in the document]
zamulla: After reading the whole guide, I am thinking that this diagram is missing the local Swift package itself: the framework is integrated with the package, and the package, in turn, is consumed by the Xcode project. We work on the local package throughout the guide, and only at the end it is imported as a dependency for the actual app, I think that should be reflected in the picture.

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -84,26 +82,63 @@ To connect the Kotlin framework to your Xcode project and use the Kotlin code in
...
+   }
+   
+   #Preview {
+       ContentView()
+   }
+   ```
+   
+7. Build the project in Xcode. If everything is set up correctly, the project build will be successful.
+   
+There's a couple more factors worth considering:
```
zamulla, line 136:
```suggestion
There are a couple more factors worth considering: 
```

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -1,8 +1,20 @@
 [//]: # (title: Using Kotlin from local Swift packages)
 
-In this tutorial, you'll learn how to add a dependency on a Kotlin framework in a local package in the Swift package manager.
-This is useful if you want to simultaneously connect your Kotlin Multiplatform module to several SPM modules
-in your Xcode project.
+In this tutorial, you'll learn how to integrate a Kotlin framework from a Kotlin Multiplatform project into a local
+package in the Swift package manager.
```
zamulla, line 4: As I see it, the end result is not integrating a framework into a package, but using Kotlin code in an Xcode project. Should we end the "you'll learn" sentence with that?

zamulla (changes requested):
I asked some questions and made some suggestions, please tell me what you think 👍

topics/development/multiplatform-spm-local-integration.md
```diff
@@ -1,8 +1,20 @@
...
+
+This is a local integration method that can work for you if:
```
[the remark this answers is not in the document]
danil-pavlov: This doesn't really work without a verb. Plus, here, we just state the type of the integration method

topics/development/multiplatform-spm-local-integration.md
danil-pavlov: Yes, we need to distinguish this particular local integration method from the standard one provided by the wizard. That's why this part is at the beginning

topics/development/multiplatform-spm-local-integration.md
danil-pavlov: The negative highlights the profit here, you don't need to rebuild both projects compared to the wizard approach. the second sentence answers other questions

images/migrate-app/direct-integration-scheme.svg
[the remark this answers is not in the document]
danil-pavlov: Removed the border, thanks. It was a leftover from the previous version. The diagram itself is a general direct integration illustration, which still works here because the package is a part of the Xcode project

If(when) we have both direct integration tutorials next to each other, we can discuss with the devs how to better visualize the distinction

topics/development/multiplatform-spm-local-integration.md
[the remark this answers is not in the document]
danil-pavlov: The problem here is the clash in capitalization styles, with Apple writing every word in the upper-case, including all UI elements. I think it's questionable whether we should follow. I've added the acronym to be clear

danil-pavlov: The end goal is actually to set up the integration, the code example is to show (and check) that everything works

topics/development/multiplatform-spm-local-integration.md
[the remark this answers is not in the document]
zamulla: I missed the "that" in my suggestion, corrected that. But that's minor, do as you see fit.

About "state the type of the integration method" I want say that even inside JetBrains we had a lot of trouble defining these types and calling them something. So merely stating the type we came up with may not be enough. This may not be an immediate concern for a hidden single-purpose article, but consider this going forward.

topics/development/multiplatform-spm-local-integration.md
zamulla: My question is quite literal: if you don't need to build both projects, which one do you need to build? It may be less ambiguous to change this to "...you only need to build the Kotlin Multiplatform project to see changes made in common code reflected in Swift, whereas the build phase approach requires rebuilding the iOS project as well."

topics/development/multiplatform-spm-local-integration.md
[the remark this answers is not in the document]
danil-pavlov: I agree, I think the overview of all the integration methods should help clear things up (and possibly moving this page closer to the overview and other methods)

topics/development/multiplatform-spm-local-integration.md
danil-pavlov: OK so, to address these concerns, I decided to start from scratch and revamp these two paragraphs. The first is now about how users will set up and work with this method, and the second is the comparison with the other one

topics/development/multiplatform-spm-local-integration.md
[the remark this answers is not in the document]
zamulla: I mean, capitalization it's still part of the name, like with "npm", for example. But I don't think it will actually be confusing, the acronym is the search term and should help.

topics/development/multiplatform-spm-local-integration.md
zamulla: Yes, I think it's much more to the point now, thanks!

topics/development/multiplatform-spm-local-integration.md
zamulla: Maybe I'm going too abstract here, and it's another point that is more about the integration methods overview than this particular guide.

PR #203, opened 2024-08-27
fix: compatibility for older versions of Kotlin

zamulla:
We removed the table and other content related to compatibility between Kotlin and Compose compiler, but there are enough users with older Kotlin versions who are left in the dark about this.

### DIFF
--- a/topics/compose/compose-compatibility-and-versioning.md
+++ b/topics/compose/compose-compatibility-and-versioning.md
@@ -62,11 +62,13 @@ When a new version of Jetpack Compose is released, we:
 
 The gap between a Compose Multiplatform release and a Jetpack Compose release is usually 1–3 months.
 
-### Early access versions of Compose Multiplatform
+### Development versions of Compose Multiplatform
 
-Early access versions of the Compose Multiplatform compiler plugin, like `1.7.0-alpha03`,
-are not available in [Maven Central](https://central.sonatype.com/).
-To use early access versions, add this line to your list of repositories:
+Development versions of the Compose Multiplatform compiler plugin (for example, `1.7.0-dev1776`) are built without a set schedule,
+to test updates between formal releases.
+
+These builds are not available in [Maven Central](https://central.sonatype.com/).
+To access them, add this line to your list of repositories:
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


topics/compose/compose-compatibility-and-versioning.md
```diff
@@ -62,11 +62,24 @@ When a new version of Jetpack Compose is released, we:
 
 The gap between a Compose Multiplatform release and a Jetpack Compose release is usually 1–3 months.
 
-When you build your application for Android, the artifacts published by Google are used. For example, if you apply the
-Compose Multiplatform 1.5.0 Gradle plugin and add `implementation(compose.material3)` to your `dependencies`, then your
+### Early access versions of Compose Multiplatform
+
+Early access versions of the Compose Multiplatform compiler plugin, like `1.7.0-alpha03`,
```
igordmn, line 67:
```suggestion
Early access versions of the Compose Multiplatform compiler plugin, like `1.7.0-dev1756`,
```
We publish alpha/beta to Maven Central. Only dev's aren't published.

igordmn (approved):
Besides one comment, lgtm

topics/compose/compose-compatibility-and-versioning.md
zamulla: Can we call them something? I got confused by "early access" in the existing text, and other people probably will, too. "Development versions" sounds too vague. "Nightly versions", "snapshot versions"?

igordmn: Either `weekly versions`, or `development versions`. I am in favour of the latest, because `weekly` isn't true, we can often publish it more often manually.

Also, we most probably change versioning soon (in 1-3 weeks). It will be not `1.7.0-dev1756`, it will be `1.7.0-alpha03+build1756` (when we release `1.7.0-alpha03`, we start build `1.7.0-alpha04+build1766`)

zamulla: If there isn't any regularity, let's use "development versions" for now.

I'll be watching for the new builds, as soon as the new template hits I'll update it and explain it.

ElviraMustafina (approved):

PR #208, opened 2024-08-29
feat: reasons to adopt Kotlin Multiplatform

Daria-Voronina:
This is a new article about reasons to adopt Kotlin Multiplatform. The article describes ten reasons to adopt KMP featuring examples and case studies with references to other relevant docs.


koshachy (approved):
Great job!
I've slightly updated your PR and cleaned up the unnecessary pics.

PR #222, opened 2024-09-06
fix: missed the direct configuration option

zamulla:


topics/compose/compose-compiler.md
```diff
@@ -273,6 +273,12 @@ To disable a feature flag that is enabled by default, call the `disabled()` func
 featureFlags = setOf(ComposeFeatureFlag.StrongSkipping.disabled())
 ```
 
+If you are configuring the Compose compiler directly, use the following syntax to pass feature flags to it:
+
+```
```
sarahhaggarty, line 278:
```suggestion
```none
```
This makes the code render as a code block rather than just inline code.

sarahhaggarty (approved):
Just one suggestion to fix some formatting. LGTM! 🚀

PR #231, opened 2024-09-16
feat: new Swing interop page

ElviraMustafina:
- new page for Swing interop
- new gifs
- minor updates to code samples: added titles for windows, removed unused imports


topics/compose/compose-desktop-swing-interoperability.md
```diff
@@ -0,0 +1,517 @@
...
+The main goals of interoperability between Compose Multiplatform and Swing include:
+* Simplifying and smoothing the migration process of Swing applications to Compose Multiplatform.
+* Enhancing Compose Multiplatform applications using Swing components when no Compose analogues are available.
+
+In many cases, it's more effective to implement a missing component directly in Compose Multiplatform (and contribute it to the community) rather than using a Swing component within a Compose Multiplatform application.
+
+## Swing interop use cases and limitations
+
+When integrating Compose Multiplatform with Swing, keep in mind that these two technologies have different approaches to content rendering.
+Compose Multiplatform renders all its content using a single heavyweight Swing component and relies on logical rendering layers, while Swing operates using both heavyweight and lightweight components (`Component`/`JComponent`).
```
m-sasha, line 14: > while Swing operates using both heavyweight and lightweight components (`Component`/`JComponent`)

That's neither correct, nor very important. 
- Swing actually has the same approach as Compose - they both render all their content on a single "system" (heavyweight) component. Swing's "lightweight" components roughly correspond to Compose element nodes, in the sense that they are virtual, rather than system, components. The difference is that Compose renders using a different rendering engine (Skia), but I don't think that's important here.
- `Component` is not heavyweight. Some of its subclasses are heavyweight.
- Swing mostly operates using lightweight components, with a single heavyweight (the window) in the root.

topics/compose/compose-desktop-swing-interoperability.md
```diff
@@ -0,0 +1,517 @@
...
+In Swing logic, Compose Multiplatform is just another heavyweight component, and Swing interacts with it similarly to how it interacts with other Swing components.
+
+The first use case involves adding a Compose Multiplatform component to a Swing application. You can achieve it using the `ComposePanel` Swing component to render the Compose Multiplatform part of the application. 
+From Swing's perspective, `ComposePanel` is another Swing component that handles it accordingly.
+Note that all Compose Multiplatform components, including popups, tooltips, and context menus, will be rendered within Swing's `ComposePanel` and positioned and resized inside it. 
+Therefore, consider replacing these components with Swing-based implementations.
```
m-sasha, line 20: We have a new, experimental mode, where popups (and tooltips, context menus) are created as system windows. It's worthwhile to document it. @MatkovIvan

topics/compose/compose-desktop-swing-interoperability.md
m-sasha: We also have an experimental mode where Compose is rendered on a Swing component directly, which improves the layering of Swing and Compose elements. @MatkovIvan

m-sasha (changes requested):
We should also document our new flags:
- `compose.swing.render.on.graphics`
- `compose.layers.type`
- `compose.interop.blending`

topics/compose/compose-desktop-swing-interoperability.md
MatkovIvan: See [Render to swing graphics](https://blog.jetbrains.com/kotlin/2023/08/compose-multiplatform-1-5-0-release/#enhanced-swing-interop)

topics/compose/compose-desktop-swing-interoperability.md
MatkovIvan: See [Popup/Dialog as a separate view or window](https://www.jetbrains.com/help/kotlin-multiplatform-dev/whats-new-compose.html#desktop-experimental)

topics/compose/compose-desktop-swing-interoperability.md
ElviraMustafina: Added "Experimental off-screen rendering﻿" and "Experimental separate views for popups"

topics/compose/compose-desktop-swing-interoperability.md
ElviraMustafina: Added "Experimental interop blending"

topics/compose/compose-desktop-swing-interoperability.md
```diff
@@ -163,6 +178,98 @@ fun Counter(text: String, counter: MutableState<Int>) {
...
+
+> Off-screen rendering is [Experimental](supported-platforms.md#compose-multiplatform-ui-framework-stability-levels),
+> and you should use it only for evaluation purposes.
+>
+{style="warning"}
+
+To enable off-screen rendering, use the `compose.swing.render.on.graphics` system property:
+
+```kotlin
+System.setProperty("compose.swing.render.on.graphics", "true")
```
m-sasha, line 196: It's important to set this at the start of your application, before calling any Compose code. It's even better to set it via the command that runs the JVM by passing `-Dcompose.swing.render.on.graphics=true` to it.

m-sasha (approved):

topics/compose/compose-desktop-swing-interoperability.md
```diff
@@ -0,0 +1,660 @@
...
+By default, the interop view implemented using the `SwingPanel` wrapper is always rectangular and in the foreground, 
+on top of any Compose Multiplatform components. To make popup elements easier to use, we introduced experimental support for 
+interop blending.
+
+> Interop blending is [Experimental](supported-platforms.md#compose-multiplatform-ui-framework-stability-levels),
+> and you should use it only for evaluation purposes.
+>
+{style="warning"}
+
+To enable this experimental feature, use the `compose.interop.blending` system property:
```
MatkovIvan, line 466: Please copy a note about "The property must be set before executing any Compose code in your application" here too

MatkovIvan (approved):

zamulla (changes requested):
Nice!

I think we could make the purpose of the article a bit clearer, left some suggestions about that.

topics/compose/compose-desktop-swing-interoperability.md
ElviraMustafina: @MatkovIvan Should we also recommend to set it via the command line?

MatkovIvan: It's a valid option, but the main point is that it should be set before Compose.
Not sure if we need to describe all ways to set it for each flag. This part is up to you

ElviraMustafina: Thanks! I added info about both options, just in case.

zamulla (approved):
All good, thank you! The only minor thing is that code comments should be written in third person according to the Kotlin guidelines, but no notes otherwise.

PR #250, opened 2024-10-02
Update multiplatform-ktor-sqldelight.md

kpgalligan:
Extended the section on sqlite linker flags. While I understand this is an overview for simplicity, this section is showing up as the "official solution", and there are real reasons you wouldn't want a dynamic framework.

OK if you don't want to complicate the docs, but had to try :)

### DIFF
--- a/topics/development/multiplatform-ktor-sqldelight.md
+++ b/topics/development/multiplatform-ktor-sqldelight.md
@@ -933,10 +933,14 @@ The Kotlin Multiplatform wizard generates an iOS project that is already connect
 is exported with the name specified in the `shared/build.gradle.kts` file (`baseName = "Shared"`), and imported
 using a regular `import` statement: `import Shared`.
 
-### Turn off static linking
+### Linking sqlite
+
+On iOS, the Xcode build tools need to "link" to the system sqlite library. There are two basic ways to do that.
+
+#### Turn off static linking
 
 The Kotlin Multiplatform wizard generates projects set up for static linking of iOS frameworks.
-To use the SQLDelight library, allow the iOS framework to link dynamically.
+The pros and cons of static vs dynamic linking are only important for production apps. To make linking simple for now, link your framework dynamically.
 
 To do that, open the `shared/build.gradle.kts` file and change the `isStatic` property of the `iosTarget.binaries.framework`
 block to `false`:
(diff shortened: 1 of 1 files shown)


topics/development/multiplatform-ktor-sqldelight.md
```diff
@@ -933,10 +933,14 @@ The Kotlin Multiplatform wizard generates an iOS project that is already connect
...
-### Turn off static linking
+### Linking sqlite
+
+On iOS, the Xcode build tools need to "link" to the system sqlite library. There are two basic ways to do that.
+
+#### Turn off static linking
 
 The Kotlin Multiplatform wizard generates projects set up for static linking of iOS frameworks.
-To use the SQLDelight library, allow the iOS framework to link dynamically.
+The pros and cons of static vs dynamic linking are only important for production apps. To make linking simple for now, link your framework dynamically.
```
rocketraman, line 943: These docs are a bit out of date for another reason: in recent versions of KMP, dynamic linking is already the default. So it makes no sense to have a section titled "Turn off static linking".

zamulla: What do you consider "recent versions of KMP"? Most of the samples in the docs are based on the [KMP wizard](https://kmp.jetbrains.com/), and the latest project generated there still has `isStatic = true` in that section.

rocketraman: I'm talking about [the default](https://github.com/JetBrains/kotlin/blob/58750cdaf34f113be180f9839fd982250da69f16/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/targets/native/NativeBinaries.kt#L319) for the Kotlin gradle plugin, not the KMP wizard.

zamulla: I see. The section is explicitly written in the context of the tutorial and the wizard, so I don't think there should be any confusion here.

It was pointed out to us that one can remove the `isStatic` field rather than set it to `false` since this is the default value, and maybe that's what we should advise here. I think we opted for clarity last time, but I'm not so sure now.

rocketraman: Ah, given this is written in the context of the wizard which sets it to true, it makes sense.

PR #253, opened 2024-10-07
update: add the -lsqlite3 linking flag for a static project

zamulla:
Born from this suggestion: https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/250

### DIFF
--- a/topics/development/multiplatform-ktor-sqldelight.md
+++ b/topics/development/multiplatform-ktor-sqldelight.md
@@ -942,7 +942,7 @@ To do that:
 
 1. Right-click the `iosApp/iosApp.xcodeproj` folder in Android Studio and select the **Open In** | **Xcode** menu item.
 2. In Xcode, double-click the project in the Project navigator to open its settings.
-3. Switch to the **Build Settings** tab and search for **Other Linker Flags** field.
+3. Switch to the **Build Settings** tab and search for the **Other Linker Flags** field.
 4. Double-click the value of the field, click **+** and add the `-lsqlite3` string to the list.
 
 ### Prepare a Koin class for iOS dependency injection
--- a/topics/development/multiplatform-ktor-sqldelight.md
+++ b/topics/development/multiplatform-ktor-sqldelight.md
@@ -54,7 +54,7 @@ You will use the following multiplatform libraries in the project:
 ## Add Gradle dependencies
 
 To add a multiplatform library to the shared module, you need to add dependency instructions (`implementation`)
-to the `dependencies{}` block of the relevant source sets in the `build.gradle.kts` file.
+to the `dependencies {}` block of the relevant source sets in the `build.gradle.kts` file.
 
 Both the `kotlinx.serialization` and SQLDelight libraries also require additional configuration.
 
@@ -112,7 +112,7 @@ Change or add lines in the version catalog in the `gradle/libs.versions.toml` fi
    ![Synchronize Gradle files](gradle-sync.png)
 
 5. At the very beginning of the `shared/build.gradle.kts` file, add the following lines to the
-   `plugins{}` block:
+   `plugins {}` block:
 
    ```kotlin
    plugins {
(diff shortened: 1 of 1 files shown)


topics/development/multiplatform-ktor-sqldelight.md
```diff
@@ -933,31 +933,17 @@ The Kotlin Multiplatform wizard generates an iOS project that is already connect
 is exported with the name specified in the `shared/build.gradle.kts` file (`baseName = "Shared"`), and imported
 using a regular `import` statement: `import Shared`.
 
-### Turn off static linking
+### Add the dynamic linking flag for SQLDelight
 
 The Kotlin Multiplatform wizard generates projects set up for static linking of iOS frameworks.
-To use the SQLDelight library, allow the iOS framework to link dynamically.
+To use the native SQLDelight driver on iOS, add the linker flag that would allow this particular library to be linked dynamically.
```
kpgalligan, line 939: The change is good. I think the wording will be confusing, though. Yes, sqlite is a dynamically linked library, but I had to read the sentence a couple times to understand the meaning. Most new KMP users have little-to-no native dev experience and no understanding of linking. When discussing the sqlite library, it is a dynamically linked library, but the subject the dev is immediately concerned with is the Kotlin-generated framework.

So, maybe a slight tweak. From: 

> To use the native SQLDelight driver on iOS, add the linker flag that would allow this particular library to be linked dynamically.

To:

> To use the native SQLDelight driver on iOS, add the linker flag that would allow Xcode tooling to find the system-provided sqlite library.

Alternatives:

> To use the native SQLDelight driver on iOS, add the linker flag that would allow Xcode tooling to find the system-provided sqlite binary.

Or:

> To use the native SQLDelight driver on iOS, add the linker flag that would allow Xcode tooling to find the system-provided sqlite library (which is dynamically linked).

The last one is too much info at this stage, but if you wanted to be clear that sqlite won't be added to the Kotlin-generated framework, you could mention it.

For a tutorial, though, the less detail the better. My original change was too much detail, but I was starting to see "make your Framework dynamic" as almost the correct answer for everything, which is certainly not the case. KMP devs, if working with native targets, will sooner or later need some understanding of linking.

On that topic, while this isn't a good venue for an unrelated(ish) discussion, I am curious if there are more detailed docs that attempt to explain linking in more detail? Maybe not a full discussion of linkers and `ld` but a functional tutorial for working with Xcode and Apple build tools. If I had to pick one thing that virtually no Android dev seems to understand, and avoids learning way too long for KMP, it's linking.

zamulla: I see, that's great insight into what's important here, and I agree that less detail is better here. I think your first and second variants are the same :) but it does read more to the point.

zamulla: As for your second question, there is no documentation about linking right now, but we are considering explaining linking in more detail. If only because the `isStatic` parameter is user-facing and opaque if you're not into that kind of thing. It is opaque exactly for the reason you are describing — to explain it is to pull the thread of linking in general and Xcode/Apple nuances, and that's an undertaking. But it is definitely on the radar.

kpgalligan:
@zamulla The change looks good. GitHub didn't give me the option to "approve", which is confusing as it's requesting a review, but GitHub has lots of rules.

topics/development/multiplatform-ktor-sqldelight.md
kpgalligan: > If only because the isStatic parameter is user-facing and opaque if you're not into that kind of thing. It is opaque exactly for the reason you are describing — to explain it is to pull the thread of linking in general and Xcode/Apple nuances, and that's an undertaking. But it is definitely on the radar.

It's a difficult situation. For mobile teams, the people trying to set up KMP are almost always Android devs. Beyond basic tutorials, you simply can't do much without a functional understanding of Xcode/llvm/native tooling. A lot of the time I personally spend on support for our libraries ultimately winds up being an explanation of linking.

There *might* be an opportunity to collaborate here. We have the "here's how a linker works" content, although it's spread out in various places. Also, some people find their way to it, but they're not "official docs", which reduces the communicative value. If timing and interest line up, we could maybe draft/help-draft a doc?

danil-pavlov (approved):

PR #265, opened 2024-10-26
Include CMP Conferences List app in samples repo

ashdavies:
Humbly offering my playground as a sample for Kotlin/Compose multiplatform :)

### DIFF
--- a/topics/multiplatform-samples.md
+++ b/topics/multiplatform-samples.md
@@ -555,7 +555,6 @@ You can find even more sample projects on GitHub via the [kotlin-multiplatform-s
           </list>
       </td>
   </tr>
-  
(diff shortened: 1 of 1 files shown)


zamulla:
Hi! Thanks for the contribution, but we don't usually accept PRs altering the list of samples — to keep the size of the list under control, as it's hard to maintain a useful manual catalog.

However, we do recommend everyone with a good sample of KMP in action to use the [kotlin-multiplatform-sample](https://github.com/topics/kotlin-multiplatform-sample) topic on GitHub (or the [kotlin-multiplatform](https://github.com/topics/kotlin-multiplatform) topic for KMP projects in general).

Thanks again!

PR #266, opened 2024-10-29
new page for keyboard events

ElviraMustafina:
- new page for keyboard events
- new gifs with visible branding 
- updated code samples: added titles for windows, removed unused imports and experimental opt-ins, replaced deprecated `Dialog()`


igordmn (approved):

topics/compose/compose-desktop-keyboard.md
```diff
@@ -0,0 +1,170 @@
+[//]: # (title: Keyboard events)
+
+In Compose Multiplatform for desktop, you can manage keyboard events by setting up event handlers in two different scopes:
+
+* Event handlers based on the focused element
+* Event handlers in the window scope
```
zamulla, line 6: I would err on the side of "let's put a period at the end of list items here". They are not exactly complete sentences, but long enough.

topics/compose/compose-desktop-keyboard.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Events in a focused component
+
+This approach implies that a key pressed on a hardware keyboard is handled as an event for the currently focused component.
```
zamulla, line 10: Is the word "hardware" pertinent here? For example, when we discuss phone simulators and real devices, there are salient differences in how we use them; if an emulated keyboard would not make a difference here, it may only confuse people to mention hardware specifically.

zamulla, line 10: I would also rephrase away from passive voice. Maybe it's not that interesting who pressed the key, but it's nice to know the entity that handles key presses (Compose Multiplatform, probably?)

topics/compose/compose-desktop-keyboard.md
```diff
@@ -0,0 +1,170 @@
...
+
+A typical scenario is to define keyboard handlers for active controls like `TextField`. To intercept a key event before it
+triggers the default action (typing a symbol) and trigger a custom action instead, you can use both the `onKeyEvent` and `onPreviewKeyEvent` callbacks. 
+However, `onPreviewKeyEvent` is preferable for defining shortcuts, as these two modifiers differ by where the event is dispatched when
+it is not consumed: `onPreviewKeyEvent` dispatches the event to its first child, and `onKeyEvent` dispatches the event to the composable's parent.
```
zamulla, line 15: This sentence runs a bit long and introduces two different concepts: dispatch order and a recommendation for defining shortcuts. I would divide it into two — first, explain the difference between modifiers we introduced in the previous sentence, then, tie that difference to a practical application.

topics/compose/compose-desktop-keyboard.md
```diff
@@ -0,0 +1,170 @@
...
+
+The following sample demonstrates `TextField` interactions with the different actions depending on
```
zamulla, line 17:
```suggestion
The following sample demonstrates `TextField` interactions with different actions depending on
```

topics/compose/compose-desktop-keyboard.md
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+```
+{initial-collapse-state="collapsed" collapsible="true" collapsed-title="modifier = Modifier.onPreviewKeyEvent { when"}
+
+<img src="compose-desktop-key-focus.animated.gif" alt="Keyboard events in a focused component" width="600" preview-src="compose-desktop-key-focus.png"/>
+
+## Events in a window scope
+
+To define keyboard event handlers that are always active within the current window, use the `onPreviewKeyEvent` and `onKeyEvent` properties,
```
zamulla, line 76: So far we called `onPreviewKeyEvent ` and `onKeyEvent ` callbacks and modifiers, now they are also properties. While these are all correct, I think we should pick a default and use something else only when it is beneficial in context.

I guess "callback" is the best default? "Modifiers" I usually reserve for functions of the [Modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier) interface — there's a lot of them and they may be easy to confuse with something else. But tell me what you think about this use case in particular.

zamulla (changes requested):
Nice job! Left a couple of consistency suggestions.

topics/compose/compose-desktop-keyboard.md
ElviraMustafina: I divided it and moved to the second section, as it makes more sense for a bigger scope: order of the `onPreviewKeyEvent()` lambda calls allows implementing even screen-wide keyboard shortcuts.

topics/compose/compose-desktop-keyboard.md
ElviraMustafina: Then they are literally modifiers!
`fun Modifier.onPreviewKeyEvent(onPreviewKeyEvent: (KeyEvent) -> Boolean): Modifier`

igordmn: Properties in Window aren't modifiers, they are parameters:
```
Window(onPreviewKeyEvent = , onKeyEvent  = )
```

ElviraMustafina: Only in this part or everywhere? 
I refer to them as modifiers in the first section, with the example `Modifier.onPreviewKeyEvent`,
and as parameters in the second section, with the following example:
```
DialogWindow(onCloseRequest = { isDialogOpen = false },
            title = "Dialog",
            onPreviewKeyEvent = { ...
```

igordmn: Also, preferable to use `parameter`, not `property`. `Property` in Kotlin has different meaning (for example, `window.onPreviewKeyEvent = ` is a property)

igordmn: Also, better to not use `callback` in this section, because we need to emphasis that we need to pass a parameter, not a modifier which we passed in the first section.

ElviraMustafina: Thanks for the details! Then the current version is fine: no `callbacks`, no `properties`; `parameters` when passed as parameters, and `modifiers` when used with Modifier.

topics/compose/compose-desktop-keyboard.md
```diff
@@ -73,9 +72,11 @@ fun main() = singleWindowApplication (title = "Key events") {
 
 ## Events in a window scope
 
-To define keyboard event handlers that are always active within the current window, use the `onPreviewKeyEvent` and `onKeyEvent` properties,
+To define keyboard event handlers that are always active within the current window, use the `onPreviewKeyEvent` and `onKeyEvent` parameters,
 available for the `Window`, `singleWindowApplication`, and `Dialog` functions. 
-Typically, `onPreviewKeyEvent` is preferred for intercepting events.
+They differ in where the event is dispatched when it is not consumed: `onPreviewKeyEvent` dispatches the event to its first child,
```
zamulla, line 77:
```suggestion
They differ in how the event is dispatched when it is not consumed: `onPreviewKeyEvent` dispatches the event to its first child, 
```

A nitpicky edit, I just don't think a spatial metaphor is clear here.

zamulla (approved):
Thanks! One more tiny suggestion, but LGTM overall.

PR #268, opened 2024-11-03
Update onClick description to reflect the code

qurbonzoda:
The `onClick` event handler does not invoke the greeting property  with the platform name.


qurbonzoda:
Am I getting the documentation or Compose Multiplatform wrong?

zamulla (approved):
Yep, we probably missed a code change here, thanks for catching that!

PR #287, opened 2024-12-19
Compose. Support Windows ARM

igordmn:
The support was added in https://youtrack.jetbrains.com/issue/CMP-223/Finish-Windows-ARM-64-support

### DIFF
--- a/topics/whats-new/whats-new-compose-180.md
+++ b/topics/whats-new/whats-new-compose-180.md
@@ -23,9 +23,9 @@ For details, see the [Jetpack Compose documentation](https://developer.android.c
 
 ## Desktop
 
-### Support for Windows ARM64
+### Support for Windows for ARM64
 
-Compose Multiplatform introduces support for Windows ARM64, allowing you to build and run applications natively on 
+Compose Multiplatform introduces support for Windows for ARM64, allowing you to build and run applications natively on 
 ARM-based Windows devices.
 
 ## Web
--- a/topics/whats-new/whats-new-compose-180.md
+++ b/topics/whats-new/whats-new-compose-180.md
@@ -19,3 +19,10 @@ with all style preferences such as weight, width, slant, italic, custom axes, vi
 and style adaptable to specific text sizes.
 
 For details, see the [Jetpack Compose documentation](https://developer.android.com/develop/ui/compose/text/fonts#variable-fonts).
+
+## Desktop
+
+### Support for Windows ARM64
+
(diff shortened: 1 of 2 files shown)


igordmn:
Is  the branch correct for Compose 1.8.0-beta01?

zamulla:
> Is the branch correct for Compose 1.8.0-beta01?

Yup, thank you!

topics/whats-new/whats-new-compose-180.md
```diff
@@ -21,6 +21,13 @@ and style adaptable to specific text sizes.
 
 For details, see the [Jetpack Compose documentation](https://developer.android.com/develop/ui/compose/text/fonts#variable-fonts).
 
+## Desktop
+
+### Support for Windows ARM64
```
zamulla, line 26: From what I can see on microsoft.com, I don't think "Windows ARM64" is the proper term. Let's do it like this maybe? @igordmn 

```suggestion
### Support for Windows for ARM64
```

Microsoft also writes it as "Arm64", but nobody else does, so let's ignore that.

topics/whats-new/whats-new-compose-180.md
```diff
@@ -21,6 +21,13 @@ and style adaptable to specific text sizes.
...
+
+Compose Multiplatform introduces support for Windows ARM64, allowing you to build and run applications natively on
```
zamulla, line 28: And if we change it, change it here, too.

zamulla (approved):
All good, just a naming nitpick to consider.

topics/whats-new/whats-new-compose-180.md
```diff
@@ -21,6 +21,13 @@ and style adaptable to specific text sizes.
 
 For details, see the [Jetpack Compose documentation](https://developer.android.com/develop/ui/compose/text/fonts#variable-fonts).
 
+## Desktop
+
+### Support for Windows for ARM64
+
+Compose Multiplatform introduces support for Windows for ARM64, allowing you to build and run applications natively on
```
igordmn, line 28: There is a concern that "natively" can mislead people. We don't support the full Compose running natively on Windows ARM64. The main part is still runned on JVM. It is only one part (internal Skiko binary) is now compiled to native ARM64, that leads "to official supporting ARM64". Previously it was compiled to X64, so it required X64 -> ARM64 translation.

igordmn: > runned on JVM

Which is also compiled to ARM64, but we can't say that Compose itself is runned natively

PR #292, opened 2025-01-13
Remove outdated exception message

zsmb13:
The referenced libraries are now multiplatform, since 1.7.3 

https://www.jetbrains.com/help/kotlin-multiplatform-dev/whats-new-compose-170.html#new-common-modules


koshachy (approved):
Thanks!

PR #296, opened 2025-01-22
fix: mention Res import explicitly

zamulla:
In the course of expanding the resources documentation we lost the explicit mention of the way users need to import the generated class when they add Compose Multiplatform to an existing project.


topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -8,6 +8,21 @@ To regenerate the `Res` class and all the resource accessors, build the project
 
 After that, you can use the generated class to access the configured multiplatform resources from your code or from external libraries.
 
+## Importing the generated class
+
+To use the prepared resources, import the generated class, for example:
+
+```Kotlin
```
ElviraMustafina, line 15:
```suggestion
```kotlin
```

ElviraMustafina (approved):

PR #299, opened 2025-02-03
API to preload resources: docs, what's new

ElviraMustafina:


topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,114 @@ Both resource files in this example are located in the `commonMain` source set:
...
+necessary glyphs is still loading. In this case, users may temporarily see the text in default font or even empty boxes and 
+question marks instead of characters. Similarly, for images or drawables, users may observe a placeholder like a blank or 
+black box until the resource is fully loaded.
+
+To prevent visual glitches, you can preload and cache resources with the built-in browser features and Compose Multiplatform 
+preload API.
+
+### Preload resources with rel="preload"
+
+Modern browsers support resource preloading through the `<link>` tag with the `rel="preload"` attribute. This attribute allows
```
eymar, line 454: We might add this link https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload  so the reader can learn more about it.

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,114 @@ Both resource files in this example are located in the `commonMain` source set:
...
+
+```console
+   ./gradlew :composeApp:wasmJsBrowserDistribution
+```
+
+2. Find the required resource in the generated `dist` directory.
+3. Add a `<link>` tag to your `index.html` file and set the `href` attribute to the resource path:
+
+```html
+<link rel="preload" href="./composeResources/username.composeapp.generated.resources/font/FiraMono-Regular.ttf" as="fetch" type="font/ttf" crossorigin/>
```
eymar, line 470: Just want to note that this path `./composeResources/username.composeapp.generated.resources` is an example, and the actual path depends on the user project/package. The path can be looked up in the `dist` directory.

Also, maybe it's worth to mention that the tag `<link>` should be added inside `<head>` tag. 
The index html can be found in `wasmJsMain/resources` folder of the app module.

ElviraMustafina: Sure, getting the actual path was described in the previous step.
I added more details on where to insert `<link>`. Thanks!

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,115 @@ Both resource files in this example are located in the `commonMain` source set:
...
+
+The browser cache stores the raw bytes of preloaded resources that still need to be converted into a format suitable for 
+rendering, such as `FontResource` and `DrawableResource`. When the application requests the resource the first time, the 
+conversion is done asynchronously, which may again result in flickering. To further optimize the experience, Compose resources 
+have their own implicit cache for higher-level representations of the resources, that can also be preloaded.
+
+Compose Multiplatform %composeEapVersion% introduces an experimental API for preloading font and image resources on 
+web targets: `preloadFont()`, `preloadImageBitmap()`, and `preloadImageVector()`.
+
+Additionally, you can set a fallback font different from the default bundled one if you need special characters like emojis.
```
eymar, line 485: Since the example below combines all the APIs used, I was thinking maybe noting here the API for setting the fallback font would help a reader to connect the dots.

We use `fontFamilyResolver.preload` in the example below. It's not a new API, it available in earlier versions too. 

Just an idea - something simple like this could help:
"""
Additionally, you can set a fallback font different from the default bundled one if you need special characters like emojis. _The fallback font can set using `FontFamily.Resolver.preload(...)` method_
"""

Also, there could be more than one fallback font.

ElviraMustafina: Thanks for the great suggestion! I've included that in the latest commit

eymar (approved):
LGTM! thank you!

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,116 @@ Both resource files in this example are located in the `commonMain` source set:
...
+slower network connections, the fetching can cause visual glitches like a Flash of Unstyled Text (FOUT) or showing 
+placeholders instead of images.
+
+A typical example of such an issue is when a `Text()` component contains some text in a custom font, but the font containing the 
+necessary glyphs is still loading. In this case, users may temporarily see the text in default font or even empty boxes and 
+question marks instead of characters. Similarly, for images or drawables, users may observe a placeholder like a blank or 
+black box until the resource is fully loaded.
+
+To prevent visual glitches, you can preload and cache resources with the built-in browser features and Compose Multiplatform 
+preload API.
```
zamulla, line 450: It may be my impression, but by default when I'm presented by two options I assume they are alternative. When I got to the end, I realized that browser preload and CMP preload are complimentary. I think it would be nice to mention it here, so that it's clear from the start that they are designed to be used together (if I'm not making this up).

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,116 @@ Both resource files in this example are located in the `commonMain` source set:
...
+<link rel="preload" href="./composeResources/username.composeapp.generated.resources/font/FiraMono-Regular.ttf" as="fetch" type="font/ttf" crossorigin/>
+```
+
+### Preload resources with preload API
+<secondary-label ref="Experimental"/>
+
+The browser cache stores the raw bytes of preloaded resources that still need to be converted into a format suitable for 
+rendering, such as `FontResource` and `DrawableResource`. When the application requests the resource the first time, the 
+conversion is done asynchronously, which may again result in flickering. To further optimize the experience, Compose resources 
+have their own implicit cache for higher-level representations of the resources, that can also be preloaded.
```
zamulla, line 480:
```suggestion
have their own implicit cache for higher-level representations of the resources that can also be preloaded.
```
Not sure about "implicit" — maybe "internal"?

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,116 @@ Both resource files in this example are located in the `commonMain` source set:
...
+```html
+<link rel="preload" href="./composeResources/username.composeapp.generated.resources/font/FiraMono-Regular.ttf" as="fetch" type="font/ttf" crossorigin/>
+```
+
+### Preload resources with preload API
+<secondary-label ref="Experimental"/>
+
+The browser cache stores the raw bytes of preloaded resources that still need to be converted into a format suitable for 
+rendering, such as `FontResource` and `DrawableResource`. When the application requests the resource the first time, the 
+conversion is done asynchronously, which may again result in flickering. To further optimize the experience, Compose resources
```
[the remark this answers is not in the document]
zamulla: This relates to what I said above: it's not entirely clear at this point why is this _further_ optimizing the experience. Maybe we could make the connection clearer by starting the paragraph with something like "Even if you preloaded the resources in the browser, they are stored as raw bytes that still..."

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,116 @@ Both resource files in this example are located in the `commonMain` source set:
...
+conversion is done asynchronously, which may again result in flickering. To further optimize the experience, Compose resources 
+have their own implicit cache for higher-level representations of the resources, that can also be preloaded.
+
+Compose Multiplatform %composeEapVersion% introduces an experimental API for preloading font and image resources on 
+web targets: `preloadFont()`, `preloadImageBitmap()`, and `preloadImageVector()`.
+
+Additionally, you can set fallback fonts different from the default bundled option if you require special characters like emojis.
+To specify a fallback font, use the `FontFamily.Resolver.preload()` method.
+
+The following example demonstrates how to use preloading and fallback font:
```
zamulla, line 488:
```suggestion
The following example demonstrates how to use preloading and a fallback font:
```

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,116 @@ Both resource files in this example are located in the `commonMain` source set:
...
+                CircularProgressIndicator(modifier = Modifier.align(Alignment.Center))
+            }
+            println("Fonts are not ready yet")
+        }
+
+        val fontFamilyResolver = LocalFontFamilyResolver.current
+        LaunchedEffect(fontFamilyResolver, emojiFont) {
+            if (emojiFont != null) {
+                fontFamilyResolver.preload(FontFamily(listOf(emojiFont)))
+                fontsFallbackInitialized = true
```
zamulla, line 540: Let's add comments here, to make reading the code easier.

topics/whats-new/whats-new-compose-180.md
```diff
@@ -19,3 +20,19 @@ with all style preferences such as weight, width, slant, italic, custom axes, vi
...
+<secondary-label ref="Experimental"/>
+
+Compose Multiplatform introduces a new experimental API for preloading fonts and images for web targets. Preloading helps 
+prevent visual issues such as Flash of Unstyled Text (FOUT) or flickering of images and icons.
+
+The following functions are now available for loading and caching resources:
+
+* `preloadFont()` preloads fonts
+* `preloadImageBitmap()` preloads bitmap images
+* `preloadImageVector()` preloads vector images
```
zamulla, line 36: Should we mention the custom fallback font here, too, or it wasn't introduced with this release?

zamulla (changes requested):
Thanks! I wrote a lot of words in the comments, but essentially it's a little structural adjustment and some wording issues.

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -435,6 +435,116 @@ Both resource files in this example are located in the `commonMain` source set:
...
+
+A typical example of such an issue is when a `Text()` component contains some text in a custom font, but the font containing the 
+necessary glyphs is still loading. In this case, users may temporarily see the text in default font or even empty boxes and 
+question marks instead of characters. Similarly, for images or drawables, users may observe a placeholder like a blank or 
+black box until the resource is fully loaded.
+
+To prevent visual glitches, you can preload and cache resources with the built-in browser features and Compose Multiplatform 
+preload API.
+
+### Preload resources with rel="preload"
```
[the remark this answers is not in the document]
eymar: > So I'd suggest something like "Preload resources using browser features" and "Preload resources using the Compose Multiplatform preload API", to highlight the difference.

just want to add that both options are not mutually exclusive - they can be used together.

topics/compose/compose-multiplatform-resources-usage.md
eymar: here is the code with the comments:

```kotlin
@OptIn(ExperimentalComposeUiApi::class, ExperimentalResourceApi::class, InternalComposeUiApi::class)
fun main() {
    configureWebResources {
        // To override where the resource should be found
        resourcePathMapping { path -> "./$path" }
    }
    CanvasBasedWindow("Resources + K/Wasm") {
        val font1 by preloadFont(Res.font.Workbench_Regular)
        val font2 by preloadFont(Res.font.font_awesome, FontWeight.Normal, FontStyle.Normal)
        val emojiFont = preloadFont(Res.font.NotoColorEmoji).value
        var fontsFallbackInitialized by remember { mutableStateOf(false) }

        // The "root content" Composable. It or its children use the above resource.
        UseResources()

        if (font1 != null && font2 != null && emojiFont != null && fontsFallbackInitialized) {
            println("Fonts are ready")
        } else {
            // The app is not useful while the loading has not yet completed, so we show the progress indicator 
            Box(modifier = Modifier.fillMaxSize().background(Color.White.copy(alpha = 0.8f)).clickable {  }) {
                CircularProgressIndicator(modifier = Modifier.align(Alignment.Center))
            }
            println("Fonts are not ready yet")
        }

        val fontFamilyResolver = LocalFontFamilyResolver.current
        LaunchedEffect(fontFamilyResolver, emojiFont) {
            if (emojiFont != null) {
                // Preload a fallback font containing the emojis to ensure that any missing glyphs,
                // such as emojis not supported by the bundled font, are rendered correctly.
                fontFamilyResolver.preload(FontFamily(listOf(emojiFont)))
                fontsFallbackInitialized = true
            }
        }
    }
}
```
___

Just to explain why `fontFamilyResolver.preload(FontFamily(listOf(emojiFont)))` is needed and why `preloadFont(Res.font.NotoColorEmoji)` is not enough for emojis:


When a developer wants to show Text(...) using some specific font, they can simply use `Res.font.font_awesome`. But if the Text(...) contains usual text and emojis, then the emoji glyphs will be searched in `font_awesome`, which doesn't have them.
That's why we configure a fallback font (1 or more) for such cases. So when a glyph is not found in `font_awesome`, the font resolver will try to find a glyph in one of the fallback fonts.

topics/compose/compose-multiplatform-resources-usage.md
```diff
@@ -471,21 +471,21 @@ For example, to enable in-browser preloading of a font:
...
+### Preload resources using the Compose Multiplatform preload API
 <secondary-label ref="Experimental"/>
 
-The browser cache stores the raw bytes of preloaded resources that still need to be converted into a format suitable for 
-rendering, such as `FontResource` and `DrawableResource`. When the application requests the resource the first time, the 
-conversion is done asynchronously, which may again result in flickering. To further optimize the experience, Compose resources 
-have their own implicit cache for higher-level representations of the resources, that can also be preloaded.
+Even if you preloaded the resources in the browser, they are cached as raw bytes that still need to be converted into a 
+format suitable for rendering, such as `FontResource` and `DrawableResource`. When the application requests the resource 
+the first time, the conversion is done asynchronously, which may again result in flickering. To further optimize the experience,
```
zamulla, line 479:
```suggestion
for the first time, the conversion is done asynchronously, which may again result in flickering. To further optimize the experience, 
```

I think we need a preposition here.

zamulla (approved):
Thanks for the update, good to go 👍

PR #302, opened 2025-02-06
update: updating Multiplatform Journal

Daria-Voronina:
Adding visual call to actions to the journal articles, removing information about Fleet in the journal.


zamulla (approved):
👍

PR #315, opened 2025-03-13
Update multiplatform-update-ui.md

ChawkiOS:
Xcode instead of Android Studio


zamulla:
Hi! The intent in the phrase you edited is to point the user towards Android Studio, exactly because they don't have to build the project in Xcode. Did that not work for you, or the flow wasn't clear?

PR #318, opened 2025-03-20
update: update kotlin-multiplatform-flutter.md

Daria-Voronina:
Update links and add internal links from other docs.


zamulla (approved):
One syntax suggestion, otherwise good 👍

PR #322, opened 2025-03-21
update: additional breaking change for delayed coroutines in tests

zamulla:
From this PR: https://github.com/JetBrains/compose-multiplatform-core/pull/1550

### DIFF
--- a/topics/whats-new/whats-new-compose-180.md
+++ b/topics/whats-new/whats-new-compose-180.md
@@ -75,7 +75,7 @@ Because of that, this test, for example, would hang indefinitely:
 
 ```kotlin
 @Test
-fun loopInLaunchedEffectTest() = runSkikoComposeUiTest {
+fun loopInLaunchedEffectTest() = runComposeUiTest {
     setContent {
         LaunchedEffect(Unit) {
             while (true) {
@@ -109,7 +109,7 @@ waitForIdle()
 
 // Since waitForIdle no longer waits for the delayed LauchedEffect() to complete,
 // the test needs to advance time to make the following assertion correct
-mainClock.advanceTimeBy(1000)
+mainClock.advanceTimeBy(1001)
 
 assertEquals("1", text)
 ```
--- a/topics/whats-new/whats-new-compose-180.md
+++ b/topics/whats-new/whats-new-compose-180.md
@@ -68,6 +68,55 @@ composable(Destinations.Followers.route) { navBackStackEntry ->
 }
 ```
 
+### New handling of coroutine delays in tests
+
+Previously, Compose Multiplatform tests would not consider side effects with `delay()` calls idle.
+Because of that, this test, for example, would hang indefinitely:
+
+```kotlin
+@Test
+fun loopInLaunchedEffectTest() = runSkikoComposeUiTest {
+    setContent {
+        LaunchedEffect(Unit) {
+            while (true) {
+                delay(1000)
+                println("Tick")
+            }
+        }
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


topics/whats-new/whats-new-compose-180.md
```diff
@@ -68,6 +68,55 @@ composable(Destinations.Followers.route) { navBackStackEntry ->
...
 ```
 
+### New handling of coroutine delays in tests
+
+Previously, Compose Multiplatform tests would not consider side effects with `delay()` calls idle.
+Because of that, this test, for example, would hang indefinitely:
+
+```kotlin
+@Test
+fun loopInLaunchedEffectTest() = runSkikoComposeUiTest {
```
m-sasha, line 78: No need for "Skiko".

topics/whats-new/whats-new-compose-180.md
```diff
@@ -68,6 +68,55 @@ composable(Destinations.Followers.route) { navBackStackEntry ->
...
+setContent {
+    LaunchedEffect(updateText) {
+        if (updateText) {
+            delay(1000)
+            text = "1"
+        }
+    }
+}
+updateText = true
+waitForIdle()
```
m-sasha, line 108: What is this for?

topics/whats-new/whats-new-compose-180.md
```diff
@@ -68,6 +68,55 @@ composable(Destinations.Followers.route) { navBackStackEntry ->
...
+
+// Since waitForIdle no longer waits for the delayed LauchedEffect() to complete,
+// the test needs to advance time to make the following assertion correct
+mainClock.advanceTimeBy(1000)
```
m-sasha, line 112: I would do 1001, just in case.

topics/whats-new/whats-new-compose-180.md
zamulla: I copied the example from release notes: https://github.com/JetBrains/compose-multiplatform-core/pull/1550

"...add `mainClock.advanceTimeBy(1000)` after `waitForIdle()`, because `waitForIdle` no longer waits for the `delay`-ed coroutine to complete."

Should it be something different?

m-sasha: The point was that the waitForIdle will no longer advance the clock in this scenario because the state is considered idle before `delay` returns. So the developer needs to advance the clock manually. But then the waitForIdle is just unnecessary.

zamulla: Or in this particular case `waitForIdle()` doesn't make sense because there's nothing to wait for? So I should write `advanceTimeBy()` instead of `waitForIdle()`, not after.

m-sasha: Hmm, actually let me check first. It's possible the waitForIdle is needed to launch the effect (or advanceClock will need to be by 1016).

m-sasha: Yes, sorry, the `waitForIdle` is kind of necessary because it takes an extra frame (16ms) to start the `LaunchedEffect`. Without it, the delay only completes after 1016ms.

topics/whats-new/whats-new-compose-180.md
```diff
@@ -68,6 +68,55 @@ composable(Destinations.Followers.route) { navBackStackEntry ->
...
+        if (updateText) {
+            delay(1000)
+            text = "1"
+        }
+    }
+}
+updateText = true
+waitForIdle()
+
+// Since waitForIdle no longer waits for the delayed LauchedEffect() to complete,
```
ElviraMustafina, line 110:
```suggestion
// Since waitForIdle() no longer waits for the delayed LaunchedEffect() to complete,
```

PR #325, opened 2025-03-24
feat: programming languages cross platform

Daria-Voronina:
New article about programming languages for cross-platform development.


images/journal/kotlin-quote.png
ElviraMustafina, line 1: I can't find where this image is supposed to be used. 
Anyway, I don't think the text on the transparent background will be readable in a dark theme.

Daria-Voronina: Hi! Oh, I'm sorry, I actually removed it from the doc. And when I built the page, the image wasn’t there. I’m not sure why it’s still in the Files on GitHub.

ElviraMustafina (approved):

PR #326, opened 2025-04-03
update: note on older AGP version

zamulla:
Some users automatically upgrade AGP when Android Studio suggests it, and then the wizard-generated project may not work anymore. Since this will continue to happen until KMP supports the latest AGP, a warning about this doesn't seem out of place.

### DIFF
--- a/topics/multiplatform-onboard/multiplatform-create-first-app.md
+++ b/topics/multiplatform-onboard/multiplatform-create-first-app.md
@@ -48,8 +48,9 @@ The following instructions assume that you have all software necessary for the p
 
    Android Studio detects that the folder contains a Gradle build file and opens the folder as a new project.
 
-   > Since the Android Gradle plugin support in Kotlin Multiplatform is a little [behind the latest version](https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#version-compatibility)
-   > it is **not** recommended to automatically upgrade AGP whenever Android Studio suggests it.
+   > Android Studio may automatically suggest upgrading the Android Gradle plugin in the project to the latest version.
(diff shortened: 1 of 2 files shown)


topics/multiplatform-onboard/multiplatform-create-first-app.md
```diff
@@ -48,6 +48,11 @@ The following instructions assume that you have all software necessary for the p
 
    Android Studio detects that the folder contains a Gradle build file and opens the folder as a new project.
 
+   > Since the Android Gradle plugin support in Kotlin Multiplatform is a little [behind the latest version](https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#version-compatibility)
+   > it is **not** recommended to automatically upgrade AGP whenever Android Studio suggests it.
```
ElviraMustafina, line 52: "a little behind" feels too informal to me.
What do you think about this version:

> We aim to support the latest version of the Android Gradle plugin, however, immediate alignment is not guaranteed. We recommend avoiding automatic upgrades whenever Android Studio suggests one. Please ensure compatibility before updating.

zamulla: I would be careful with our aims here, not sure if it's even technically possible :)

But I generally rethought the structure of the note: our point here is to help people not break the tutorial, not fit the whole infrastructural context in a sentence. See https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/326/commits/4fcd867f9e344c33788e1f578ab9f5b1b9f48bb0

ElviraMustafina: Is there a tiny chance that the latest version can be supported?
Consider smth like "We don't recommend upgrading, as Kotlin Multiplatform may not be compatible with the latest AGP version"

PR #335, opened 2025-04-11
upd: resources in androidLibrary

ElviraMustafina:


zamulla (changes requested):
All good, but I'd like to clarify the intro.

zamulla (approved):
Thanks! Now I get it :)

PR #336, opened 2025-04-14
RTL languages

ElviraMustafina:


mazunin-v-jb (approved):
lgtm

topics/compose-rtl.md
```diff
@@ -0,0 +1,285 @@
...
+according to the system's locale settings when RTL languages are used.
+
+## Layout mirroring
+
+When the system locale is configured for an RTL language, Compose Multiplatform automatically mirrors most UI components.
+Adjustments include changes to paddings, alignments, and component positions:
+
+* **Paddings, margins, and alignments**  
+   Default paddings and alignments are reversed. For example, <br/>
+ `Modifier.padding(start: Dp, top: Dp, end: Dp, bottom: Dp)` will automatically adjust to
```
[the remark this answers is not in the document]
zamulla: Also, should we make all of the arguments different here? I had trouble seeing the difference at first. Something like `Modifier.padding(start: Dp = 2.dp, top: 5.dp, end: Dp = 10.dp, bottom: 8.dp)`. Then it's easier to track that start and end change places while top and bottom don't.

topics/compose-rtl.md
```diff
@@ -0,0 +1,285 @@
...
+* **Button positioning**  
+   Common UI patterns, such as the position of **Cancel** and **Confirm** buttons, adjust to RTL expectations.
+
+## Forcing layout direction
+
+You may need to keep the original orientation of some UI elements, such as logos or icons, regardless of layout direction.
+You can explicitly set the layout direction for an entire app or individual components, 
+overriding the system's default locale-based layout behavior.
+
+To exclude an element from automatic mirroring, use `LayoutDirection.Rtl` or `LayoutDirection.Ltr` to enforce a specific orientation:
```
zamulla, line 33: I think we should choose either "to exclude" or "to enforce" here, they mean roughly the same thing.

topics/compose-rtl.md
```diff
@@ -0,0 +1,285 @@
...
+CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Ltr) {
+    Column(modifier = Modifier.fillMaxWidth()) {
+        // Components in this block will be laid out left-to-right
+        Text("LTR Latin")
+        TextField("Hello world\nHello world")
+    }
+}
+```
+
+The `CompositionLocalProvider()` ensures that the specified layout direction applies to all child components in the composition.
```
zamulla, line 45: I would put this in a code block comment, then there's no need to refer to the function from another paragraph.

topics/compose-rtl.md
```diff
@@ -0,0 +1,285 @@
...
+
+### BiDi text
+
+Compose Multiplatform uses the [Unicode Bidirectional Algorithm](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics)
+to manage and render bidirectional (BiDi) text, aligning punctuation and numbers.
+
+The text should be displayed in the expected visual order: punctuation and numbers aligned correctly, 
+Arabic script flows from right to left, and English is from left to right.
+
+```kotlin
```
zamulla, line 70: I'm not sure I understand what the point of the example is: does the user have to implement similar logic (custom text field for bidirectional input)? I think we need an introductory paragraph here, and maybe a comment or two on the `TextField` function — why exactly can't I use `BasicTextField` here.

zamulla (approved):
Nice! I left a couple of suggestions, most notably to explain the BiDi example; but nothing major.

topics/compose-rtl.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Forcing layout direction
+
+You may need to keep the original orientation of some UI elements, such as logos or icons, regardless of layout direction.
+You can explicitly set the layout direction for an entire app or individual components, 
+overriding the system's default locale-based layout behavior.
+
+To exclude an element from automatic mirroring and enforce a specific orientation, 
+you can use `LayoutDirection.Rtl` or `LayoutDirection.Ltr`.
+To specify a layout direction within a scope, use the `CompositionLocalProvider()`, which ensures the layout direction
```
zamulla, line 35:
```suggestion
To specify a layout direction within a scope, use the `CompositionLocalProvider()`, which ensures that the layout direction 
```

PR #349, opened 2025-04-29
feat: Manage local resource environment

ElviraMustafina:


topics/compose-resource-environment.md
```diff
@@ -0,0 +1,288 @@
+# Manage local resource environment
+
+Learn how to manage the application's resource environment to match dynamically updated resource-related settings.
+For example, you may need to handle in-app changes of language or theme.
```
zamulla, line 4: I'd take from a different angle: you may want to provide in-app settings to users; handling changes is more about implementation.

terrakok:
I'm not sure but I'd add a message on the top, that this is a workaround of the current design problem which should be solved in the compose core.

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,290 @@
...
+            }
+    
+            return LocalSystemTheme.provides(new)
+        }
+    }
+    ```
+
+## Density
+
+To change the application's `Density`, you can use the common `LocalDensity` API, supported on all platforms:
```
zamulla, line 258: Let's define resolution density if only not to be ambiguous.

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,290 @@
...
+        key(customAppDensity) {
+            content()
+        }
+    }
+}
+```
+
+## What's next?
+
+* Get more details on resource [qualifiers](compose-multiplatform-resources-setup.md#qualifiers).
```
zamulla, line 287:
```suggestion
* Get more details on [resource qualifiers](compose-multiplatform-resources-setup.md#qualifiers).
```

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,290 @@
...
+* Learn how to [localize resources](compose-rtl.md).
```
zamulla, line 288: Is the general localization link going to be available before we publish this?

zamulla (changes requested):
Some final touches.

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,290 @@
...
+        <script src="composeApp.js"></script>
+    </html>
+    ```  
+
+## Theme 
+
+Compose Multiplatform defines the current theme via `isSystemInDarkTheme()`. 
+Themes are handled differently across platforms:
+
+* Android uses `Resources.getConfiguration().uiMode and Configuration.UI_MODE_NIGHT_MASK`.
```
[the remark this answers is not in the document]
ElviraMustafina: Nope, it's used in code like this

topics/compose-resource-environment.md
ElviraMustafina: No, let's comment this out for now

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,290 @@
...
+
+    ```kotlin
+    actual object LocalAppTheme {
+        private var default: Int? = null
+        actual val current: Boolean
+            @Composable get() = (LocalConfiguration.current.uiMode and UI_MODE_NIGHT_MASK) == UI_MODE_NIGHT_YES
+    
+        @Composable
+        actual infix fun provides(value: Boolean?): ProvidedValue<*> {
+            val configuration = LocalConfiguration.current
```
zsmb13, line 217: @terrakok I still think it's incorrect to mutate this `LocalConfiguration` instead of creating a copy and providing that into the composition for the content wrapped by `LocalAppTheme`.

My suggestion is:

```suggestion
            val configuration = Configuration(LocalConfiguration.current)
```

Without this, setting a `LocalAppTheme` will affect the value of `isSystemInDarkTheme` *everywhere*, not just within the `CompositionLocalProvider`:

```
CompositionLocalProvider(LocalAppTheme provides isDarkTheme) {
    isSystemInDarkTheme() // affected
    content()
}

isSystemInDarkTheme() // also affected!
```

This makes it impossible to tell what the real system dark theme setting is at the moment, in case you want to revert to using that real value.

zamulla (approved):
👍

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,294 @@
...
+    
+            val new = when(value) {
+                true -> (configuration.uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_YES
+                false -> (configuration.uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_NO
+                null -> default!!
+            }
+            configuration.uiMode = new
+            return LocalConfiguration.provides(configuration)
+        }
+    }
```
terrakok, line 231:
```kotlin
actual object LocalAppTheme {
    actual val current: Boolean
        @Composable get() = (LocalConfiguration.current.uiMode and UI_MODE_NIGHT_MASK) == UI_MODE_NIGHT_YES

    @Composable
    actual infix fun provides(value: Boolean?): ProvidedValue<*> {
        val new = if (value == null) {
            LocalConfiguration.current
        } else {
            Configuration(LocalConfiguration.current).apply {
                uiMode = when(value) {
                    true -> (uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_YES
                    false -> (uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_NO
                }
            }
        }
        return LocalConfiguration.provides(new)
    }
}
```

topics/compose-resource-environment.md
```diff
@@ -0,0 +1,294 @@
...
+            }
+            val new = when(value) {
+                true -> SystemTheme.Dark
+                false -> SystemTheme.Light
+                null -> default!!
+            }
+    
+            return LocalSystemTheme.provides(new)
+        }
+    }
```
terrakok, line 256:
```kotlin
@OptIn(InternalComposeUiApi::class)
actual object LocalAppTheme {
    actual val current: Boolean
        @Composable get() = LocalSystemTheme.current == SystemTheme.Dark

    @Composable
    actual infix fun provides(value: Boolean?): ProvidedValue<*> {
        val new = when(value) {
            true -> SystemTheme.Dark
            false -> SystemTheme.Light
            null -> LocalSystemTheme.current
        }

        return LocalSystemTheme.provides(new)
    }
}
```

PR #354, opened 2025-05-06
Update multiplatform-samples.md: shopping-by-kmp

razaghimahdi:
Added AutomotiveApp to shopping-by-kmp — the application is now compatible with Android Automotive OS, making it available for in-car systems.


zamulla (approved):
Thanks!
Sorry for taking a long time to get to this.

razaghimahdi:
@zamulla NP man, do you mind if I ask a question? so after all that update on shopping-by-kmp which is not over yet and still working on it(these days working on backend of shopping by kotlin ktor server) can you consider it as recommended sample? if you do I can make another pull request. if you don't I totally understand. 
tnx again for your time man

PR #355, opened 2025-05-06
CMP 1.8.0

ElviraMustafina:


v.list
```diff
@@ -4,9 +4,9 @@
 
 <vars>
 
-    <var name="composeVersion" value="1.7.3" type="string"/>
+    <var name="composeVersion" value="1.8.0" type="string"/>
 
-    <var name="composeEapVersion" value="1.8.0-rc01" type="string"/>
+<!--    <var name="composeEapVersion" value="1.8.0-rc01" type="string"/>-->
 
     <var name="kotlinVersion" value="2.0.0" type="string"/>
```
zamulla, line 11:
```suggestion
    <var name="kotlinVersion" value="2.1.20" type="string"/>
```

zamulla (approved):
Let's update Kotlin as well, and good to go.

PR #360, opened 2025-05-07
Updated Shared UI tutorial with IntelliJ IDEA

sarahhaggarty:
This PR updates the Shared UI tutorial to include IntelliJ IDEA.

Still pending: Desktop and web run configuration screenshot updates.


zamulla (approved):
👍 

I would maybe make the web and desktop screenshots a bit bigger (600 instead of 500 width), but not feeling too strongly about that.

PR #370, opened 2025-05-12
feat: Implementing layouts

ElviraMustafina:


topics/compose/compose-layout.md
```diff
@@ -0,0 +1,106 @@
...
+
+Compose uses a single-pass layout model for rendering the UI tree. Here are the key layout phases:
+
+* **Measuring**. Each composable is measured, passing size constraints down the tree to its children recursively.
+* **Sizing**. Tree nodes determine their sizes based on constraints and content.
+* **Placing**. The resolved sizes and placement instructions are passed back up the tree to position the composables.
+
+See [Jetpack Compose documentation on the layout model](https://developer.android.com/develop/ui/compose/layouts/basics#model) for details.
+
+## Constrained-based layouts
```
igordmn, line 72: ConstraintLayout is Android-only

topics/compose/compose-adaptive-layouts.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Designing adaptive layouts
+
+Follow these key guidelines when designing adaptive layouts:
+
+* Prefer canonical layouts patterns like list-detail, feed, and supporting pane.
+* Maintain consistency by reusing shared styles for padding, typography, and other design elements. Keep navigation patterns 
+consistent across devices while following platform-specific guidelines.
+* Break complex layouts into reusable composables for flexibility and modularity.
+* Adjust for screen density and orientation using tools like `LocalConfiguration` and `LocalDensity`.
```
igordmn, line 14: `LocalConfiguration` is Android-only. There is no common API for orientation.

I would suggest just use `window size classes` to determine that

topics/compose/compose-adaptive-layouts.md
```diff
@@ -0,0 +1,42 @@
+# Adaptive layouts
```
igordmn, line 1: Should we just retell the same info from the android side?

After reading this, I feel that it is just rewording/reduction of the original document, with possible omitting the context, important details, and fancy pictures. For a new reader it looks less understandable without them. For an Android developer, who knows Jetpack Compose, it might difficult to know Compose Multiplatform specifics, because they should re-read the same information they already read, but with different words.

Maybe we radically reduce the content of these pages, and just have links and necessary additional details, that don't exist on the original page?

topics/compose/compose-adaptive-layouts.md
```diff
@@ -0,0 +1,42 @@
...
+
+To provide a consistent user experience on all types of devices, adapt your app's UI to different display sizes, orientations, 
+and input modes.
+
+## Designing adaptive layouts
+
+Follow these key guidelines when designing adaptive layouts:
+
+* Prefer canonical layouts patterns like list-detail, feed, and supporting pane.
```
igordmn, line 10: Without the context, it might be difficult to understand what this is. Maybe we need to add a link to it.

The original link mentions:
- that it is in the context of material theme
- images and descriptions of the layout

topics/compose/compose-adaptive-layouts.md
```diff
@@ -0,0 +1,42 @@
...
+* Maintain consistency by reusing shared styles for padding, typography, and other design elements. Keep navigation patterns 
+consistent across devices while following platform-specific guidelines.
+* Break complex layouts into reusable composables for flexibility and modularity.
+* Adjust for screen density and orientation using tools like `LocalConfiguration` and `LocalDensity`.
+
+For a detailed guide, check out [Adaptive layout do's and don'ts](https://developer.android.com/develop/ui/compose/layouts/adaptive/adaptive-dos-and-donts).
```
igordmn, line 16: Should we point to this one specific section? We already point to the main guide at the bottom: https://developer.android.com/develop/ui/compose/layouts/adaptive

topics/compose/compose-layout-modifiers.md
```diff
@@ -0,0 +1,191 @@
...
+
+```kotlin
+@Composable
+fun Card() {
+    Row(
+        // Takes the full width of its parent
+        modifier = Modifier.fillMaxWidth() 
+    ) {
+        Image(
+            model = "https://example.com/image.jpg",
```
igordmn, line 137:
```suggestion
            /*...*/
```
This function doesn't exist in Compose. I suggest write the same way as in the original doc

topics/compose/compose-adaptive-layouts.md
ElviraMustafina: There will be another section about @Preview, so the links aren`t so close to each other

topics/compose/compose-adaptive-layouts.md
ElviraMustafina: We need an entry point anyway, as well as the basic info necessary to introduce the technology to users without a Jetpack background. 
I removed some 'advanced' parts, now it looks more like a 'definition-code example-link' structure.

igordmn: This is just a recommendation from me as a reader. I see it as not helpful if it just copies the info from another place. But I have a skewed perspective anyway. You should have more context how to describe it well.

If we want to provide basic info, than my recommendation would be removing:
- "Adaptive layouts" completely, because you usually don't think about them the start of development, especially if you just started to look at the framework
- "Working with modifiers", because:
  - [the original page looks nicer](https://developer.android.com/develop/ui/compose/modifiers) with pictures
  - it is platform independent
  - it is similar to any other API we have in Compose, just more frequently used
  - it provides a list of modifiers. Copying lists is harmful comparing to copying tutorials/getting started, as it creates 2 sources of truth, often out of sync in the future

Maybe some of the info from "Working with modifiers" can be moved into basics, just to show abilities, but not as the source of truth.

igordmn: There is another option - agree with Google, copy without changes, and make a link to the original article

topics/compose/compose-adaptive-layouts.md
igordmn: My point is about that it is strange that we picked this particular section to point on. It looks random, as we have another interested sections:
image. As we have "Learn more about adaptive layouts in the", I would just remove this sentence here.

igordmn: "Adaptive layout do's and don'ts" is similar to FAQ, not a detailed guide

zamulla: I would agree, this link is too high on the page to link to best practices, before we handled the basic stuff.

topics/compose/compose-adaptive-layouts.md
ElviraMustafina: The "Working with modifiers" and "Adaptive layouts" documentation are part of the product requirements. I double-checked with our PM and got approval for the current structure.

I don't think we should change the strategy until there's progress on the CMP-Jetpack collaboration for documentation. We may eventually merge or start actively contributing to each other's docs, so let's keep things as they are for now.

topics/compose/compose-adaptive-layouts.md
```diff
@@ -0,0 +1,41 @@
...
+Follow these key guidelines when designing adaptive layouts:
+
+* Prefer [canonical layouts](https://developer.android.com/develop/ui/compose/layouts/adaptive/canonical-layouts) patterns 
+  like list-detail, feed, and supporting pane.
+* Maintain consistency by reusing shared styles for padding, typography, and other design elements. Keep navigation patterns 
+consistent across devices while following platform-specific guidelines.
+* Break complex layouts into reusable composables for flexibility and modularity.
+* Adjust for screen density and orientation.
+
+## Using window size classes
```
igordmn, line 17: We need a [code snippet](https://developer.android.com/develop/ui/compose/layouts/adaptive/use-window-size-classes#manage_layouts_with_window_size_classes) for this section on how to get `WindowSizeClass`, otherwise it will be difficult to understand how to use it.

Also it needs `implementation("org.jetbrains.compose.material3.adaptive:adaptive:adaptiveVersionFromReleaseNotes")` to be included in the project. The API is from there. Not sure why the original Jetpack doc isn't included this info. They don't have it in the generated by wizard project (I just checked that).

topics/compose/compose-adaptive-layouts.md
igordmn: Ok. I reread these sections to check if they CMP compatible and understandable without reading the original doc, added a new comment.

topics/compose/compose-adaptive-layouts.md
ElviraMustafina: Good point! I've added an example here: https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/391

Just to clarify: in the 1.7.3 what's new, we used `material3-window-size-class`:
```
commonMain.dependencies {
    implementation("org.jetbrains.compose.material3:material3-window-size-class:1.7.3")
}
```

igordmn: > material3-window-size-class

I see 2 versions of `WindowSizeClass`:
- in [window](https://github.com/androidx/androidx/blob/be7268d0ac1b9f52544e7226534fb32eb468cf27/window/window-core/src/commonMain/kotlin/androidx/window/core/layout/WindowSizeClass.kt#L23) module
- in [adaptive]() module

The first is described in the Jetpack Compose docs [here](https://developer.android.com/codelabs/basic-android-kotlin-compose-adaptive-navigation-for-large-screens#5), the second is described [here](https://developer.android.com/develop/ui/compose/layouts/adaptive/use-window-size-classes#manage_layouts_with_window_size_classes). 

Maybe the first one is just an old API, not sure. 

Who participated in the discussion of https://www.jetbrains.com/help/kotlin-multiplatform-dev/whats-new-compose-170.html#material3-material3-window-size-class? We need to discuss this issue with them.

igordmn: Overall, I think, it can be not matter much for this specific section, as we should write the same info as in the Jetpack Compose docs

ElviraMustafina: @terrakok, we need your help!

ElviraMustafina: The `org.jetbrains.compose.material3:material3-window-size-class` version cannot be used in common code. 
The `org.jetbrains.compose.material3.adaptive:adaptive` one is newer and multiplatform, so that's the correct one.

PR #374, opened 2025-05-12
Apple watchOS -> Jetpack Compose on Wear OS

ElviraMustafina:
The PeopleInSpace sample no longer supports watchOS but now works with Wear OS


zamulla (approved):
👍

PR #377, opened 2025-05-13
Update iOS stability level to Stable

YaminMahdi:
### Changes
- Updated the iOS stability level from Beta to Stable in the platform support table.

### Reason
- Reflecting the latest stability status of iOS support in Compose Multiplatform.


zamulla (approved):
👍  thanks for your thoroughness :)

PR #382, opened 2025-05-13
update: a set of updates for tutorials with the new tooling

zamulla:
* Quickstart with the general setup ✅ 
* Shared UI tutorial ✅ 
* Native UI tutorial ✅
* Multiplatform library tutorial ✅
* Testing multiplatform tutorial ✅
* Making your Android app multiplatform ✅
* Share data access layer ✅
* Hot Reload guide ✅


images/compose-create-first/compose-unresolved-references.png
evilya, line 1: I'm not sure if that `[UNRESOLVED_REFERENCE]` supposed to be visible to normal users 🤔 it might be coming from developer mode or something, but I have no idea how to disable it

zamulla: I made screenshots with Android Studio for now, but it would be nice to find out how to turn this off in IDEA :)

PR #398, opened 2025-05-19
Swap tiles in Get started

sarahhaggarty:
This PR makes the Shared UI tutorial take first place on the Get started page.


zamulla (approved):
👍

PR #403, opened 2025-05-20
feat: more SwiftUI and UIKit examples

danil-pavlov:


ElviraMustafina (commented):
I've added a few minor suggestions

ElviraMustafina (approved):
Maybe add more code comments? Ideally, like in the camera view with the UIKit example. At least for other advanced examples. 
Another note on code comments: our guidelines recommend using verbs in the third person.

PR #434, opened 2025-06-10
Fix view controller class name

woxtu:
This PR fixes the class name in the docs, because `UIViewController` doesn't seem to have a constructor that takes `UIKitInteropProperties`.


zamulla (approved):
Thanks for catching that!

PR #436, opened 2025-06-11
CMP tutorial: Small update to time instructions

sarahhaggarty:
This PR updates the CMP tutorial to clarify that all dependencies for the `todaysDate()` function need to be taken from the `kotlinx.datetime` package and not `kotlin.time`.


zamulla (approved):
👍

PR #440, opened 2025-06-13
update: more oblique Kotlin compatibility statement after K2

zamulla:


topics/compose/compose-compatibility-and-versioning.md
```diff
@@ -1,7 +1,7 @@
 [//]: # (title: Compatibility and versions)
 
 Compose Multiplatform releases ship separately from Kotlin and Jetpack Compose releases. This page contains information
-about Compose Multiplatform releases, the compatibility between different releases, and the release cycles.
+about Compose Multiplatform releases, their release cycles, and compatibility with other components.
```
ElviraMustafina: Maybe 'the release cycles'? 'Their' makes it sound like you're referring to the release cycles of CMP releases

ElviraMustafina (approved):
Just one small comment. Otherwise, LGTM

topics/compose/compose-compatibility-and-versioning.md
zamulla: I feel that neither option works very well because both of them sound like _something_ and aren't concrete.
Changed for "Compose release cycles" which doesn't explain everything, but also not going into too much detail in the first paragraph.

PR #445, opened 2025-06-20
add `onWasmReady`

ElviraMustafina:


topics/compose/compose-navigation-routing.md
```diff
@@ -243,32 +258,35 @@ The code that does the matching needs to run before the `window.bindToNavigation
...
-                    initRoute.startsWith("find_id") -> {
-                        // Parses the string to extract route parameters before navigating to it
-                        val id = initRoute.substringAfter("find_id_").toLong()
-                        navController.navigate(Id(id))
-                    }
-                    initRoute.startsWith("patient") -> {
-                        val name = initRoute.substringAfter("patient_").substringBefore("_")
-                        val id = initRoute.substringAfter("patient_").substringAfter("_").toLong()
-                        navController.navigate(Patient(name, id))
+    // `onWasmReady` ensures initialization in Kotlin/JS, not required in Kotlin/Wasm
```
eymar: to be precise, `onWasmReady` is not available in Kotlin/Wasm at all, so this snippet won't compile when copied as is and pasted in kotlin/wasm source set.

I understand that duplicating this snippet is not good. Maybe the comment could specify this distinction between k/js and k/wasm.

eymar (approved):

topics/compose/compose-navigation-routing.md
zamulla: I'm not sure which I prefer, because people can and will copy-paste this without reading and be upset when it doesn't work.

@ElviraMustafina I would separate the JS and Wasm code clearly, maybe collapsing both of the big code blocks. But tell me how you see it.

ElviraMustafina: I've added separate, switchable tabs for Kotlin/Wasm and Kotlin/JS code.

topics/compose/compose-navigation-routing.md
```diff
@@ -63,7 +63,8 @@ Users can use the **Back** and **Forward** buttons in the browser to move betwee
 as well as use the address bar to understand where they are and get to a destination directly.
 
 To bind the web app to the navigation graph defined in your common code,
-call the `window.bindToNavigation()` method in your Kotlin/Wasm or Kotlin/JS code:
+you can use the `window.bindToNavigation()` method in your Kotlin/Wasm code. For Kotlin/JS, ensure this code runs within
```
zamulla: I would expand a tiny bit here: something like "you can use the same method in Kotlin/JS, but make sure to wrap the code..."

topics/compose/compose-navigation-routing.md
```diff
@@ -236,40 +251,81 @@ to match manually entered URLs to destination routes.
...
+            ExperimentalComposeUiApi::class,
+            ExperimentalBrowserHistoryApi::class,
+            ExperimentalSerializationApi::class
         )
-    }
-}
-```
-<!--{default-state="collapsed" collapsible="true" collapsed-title="val initRoute = window.location.hash.substringAfter( ..."}-->
+        fun main() {
+            onWasmReady {
```
zamulla, line 302: It would still be great to comment on what this actually does — here, or at the top, where we mention it for the first time. Or both :)

zamulla (approved):
👍

PR #449, opened 2025-06-23
resources toc: list of links

ElviraMustafina:


topics/compose/compose-multiplatform-resources.md
```diff
@@ -14,16 +14,16 @@ When working with resources in Compose Multiplatform, consider the current condi
...
+* [](compose-multiplatform-resources-usage.md)
 
-### Manage local resource environment
+  Learn how to use the automatically generated accessors to access resources directly in your UI code.
 
-To manage the application's resource environment like in-app theme and language, 
-see the [Local resource environment](compose-resource-environment.md) tutorial.
+* [Local resource environment](compose-resource-environment.md)
+  
+  Customize your app's resource environment like in-app theme and language.
```
zamulla: I feel like the point of that page is not so much to customize your resource environment but rather allow users to switch between environments while using appropriate resources. The initial problem is not setting it up, but changing it on the fly.

PR #458, opened 2025-07-06
update iOS compose stability in FAQ

james-yun:
Was just reading through the docs and noticed that it currently mentions that `Compose Multiplatform is only Stable for Android and desktop at the moment`. iOS was recently made stable, hence, I'm updating this line.

https://www.jetbrains.com/help/kotlin-multiplatform-dev/faq.html#between-which-platforms-can-i-share-my-ui


ElviraMustafina (approved):
Oops, nice catch! Thanks for noticing that

AStout75:
Big open source contributor guy

PR #466, opened 2025-07-11
update: old instructions on CocoaPods

zamulla:


zamulla:
I think I'm done with the mentions of the old functionality _and_ exclusive Android Studio mentions in general. The only place we still use AS exclusively is the "make your app multiplatform" tutorial, which is going to be superseded in a bit by the new migration guidance.

danil-pavlov (approved):

PR #496, opened 2025-09-01
Web target updates

AlejandraPedroza:


images/development/diagrams/default-hierarchy-example.svg
JSMonk, line 1: Should we add that under the `web` are `js` and `wasmJs` targets?

bashor (approved):

images/development/diagrams/manual-hierarchical-structure.svg
zamulla, line 1: The "webMain" label in the new diagram is slightly off-center and in slightly larger font than everything else, no?

zamulla (approved):
There's only one tiny graphic artefact to fix, otherwise looks nice!

PR #514, opened 2025-09-15
Clarify artifact availability time in publishing guide

LouisCAD:
I witnessed an 8 hour long delay recently, so setting a 15-30minutes delay only might be a bit deceptive for those times.


zamulla (approved):
I think that's fair. I don't like talking about specific wait times at all here, but don't see how we can avoid that, either.

PR #517, opened 2025-09-18
chore: clearer indication of the Xcode template

zamulla:
In the "Make your app multiplatform" tutorial.


zsmb13 (approved):
Thanks!

topics/multiplatform-integrate-in-existing-app.md
```diff
@@ -352,7 +352,7 @@ business logic in it.
 ### Create an iOS project in Xcode
 
 1. In Xcode, click **File** | **New** | **Project**.
-2. Select a template for an iOS app and click **Next**.
+2. Select a template for an iOS app: go to the **iOS** tab and select the basic **App** template, then click **Next**.
```
vnikolova: This looks like 2 steps, while the screenshot only highlights one of them and the wording seems a little excessive. I suggest the following with the new screenshot:

2. In the dialog, switch to the **iOS** tab.
[new screenshot]
3. Select the **App** template, then click  **Next**.

But I think the old screenshot could work with one step, such as:
2. In the **iOS** tab, select the **App** template, then click **Next**.

Wdyt?

zamulla: I was on the fence about marking an already highlighted tab in the screenshot, but I do think that people may overlook this if there are exactly the same **App** templates in several tabs. Which means we should use the new picture.

But it did feel that this is too many instructions for one step, I adjusted it as you suggested 👍

vnikolova: Ah I didn't realise there was another template with the same name. That makes sense then!

vnikolova (approved):

PR #522, opened 2025-09-30
update: clarify the latest supported Java version for Compose Hot Reload

zamulla:


AbdullinAM (approved):

topics/tools/compose-hot-reload.md
```diff
@@ -4,7 +4,9 @@
 
 [Compose Hot Reload](https://github.com/JetBrains/compose-hot-reload) helps you visualize and experiment with UI changes while working on a Compose Multiplatform project.
 
-At the moment, Compose Hot Reload is only available when you include a desktop target in your multiplatform project. 
+At the moment, Compose Hot Reload is only available when you include a desktop target in your multiplatform project
+and use Java 21 or older (the latest version compatible with the required JetBrains Runtime).
```
sarahhaggarty:
```suggestion
Compose Hot Reload is  currently available only if your multiplatform project includes a desktop target
and uses Java 21 or earlier (the latest version compatible with the required JetBrains Runtime). 
```
What do you think about restructuring it a bit now it's a longer sentence? I also suggest "earlier" rather than "older" because I think it makes it clearer that you mean Java 21 and earlier versions.

zamulla: Yep, accepted — I only changed "uses" to "is compatible with", because that's the crux of the matter apparently.

PR #672, opened 2026-04-20
Update Android to KMP migration guide with XML to Compose skill

EmilFlach:
Added a link to the new agent skill for AI-assisted migration from XML views to Compose.


zamulla (approved):
Thanks!

PR #683, opened 2026-05-06
Multiplatform ViewModel

ElviraMustafina:


topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+
+    ```toml
+    [versions]
+    androidx-viewmodel = "2.10.0"
+    
+    [libraries]
+    androidx-lifecycle-viewmodel-compose = { module = "org.jetbrains.androidx.lifecycle:lifecycle-viewmodel-compose", version.ref = "androidx-viewmodel" }
+    androidx-lifecycle-viewmodel-navigation3 = { module = "androidx.lifecycle:lifecycle-viewmodel-navigation3", version.ref = "androidx-viewmodel" }
+    ``` 
+    > You can track changes to the multiplatform ViewModel implementation in our [What's new](https://www.jetbrains.com/help/kotlin-multiplatform-dev/whats-new-compose.html)
```
zamulla, line 31: Should be an empty line after the code block.

topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+           // ...
+       }
     }
-}
-```
-{initial-collapse-state="collapsed" collapsible="true" collapsed-title="org.jetbrains.androidx.lifecycle:lifecycle-viewmodel-compose:%org.jetbrains.androidx.lifecycle%"}
+    ```
+    {initial-collapse-state="collapsed" collapsible="true" collapsed-title="implementation(libs.androidx.lifecycle.viewmodel.compose)"}
+
+3. If you have a desktop target, add the `kotlinx-coroutines-swing` dependency as well.
```
zamulla: If this is the last list item, I would make it not a list item.

> If you have a desktop target <...>:
> 1. In the Gradle version catalog:
> 2. In the build.gradle.kts:

This way the structure also mirrors adding the dependency to commonMain earlier.

topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+     
+    See the [`Dispatchers.Main` documentation](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-main.html) for details.
+
+Dependencies may differ depending on your code-sharing approach. See [](#levels-of-code-sharing) for details.
 
 ## Using ViewModel in common code
 
-Compose Multiplatform implements the common `ViewModelStoreOwner` interface, so in general using the `ViewModel` class
+Compose Multiplatform provides a common `ViewModelStoreOwner` implementation, so using the `ViewModel` class
 in common code is not much different from Android best practices.
```
zamulla: Should probably link the Android best practices for those who start with ViewModels here.

topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+          field = MutableStateFlow(OrderUiState())
+
+      fun setQuantity(n: Int) {
+          field.update { it.copy(quantity = n, price = "$${n * 2}.00") }
+      }
+   }
+   ```
+
+    > This example uses explicit backing fields, stabilized in Kotlin 2.4.0. When using earlier versions,
+    > add the `-Xexplicit-backing-fields` compiler option or use the old backing fields pattern with `.asStateFlow()` instead.
```
zamulla: Still feel it's useful to link the explanation here https://kotlinlang.org/docs/whatsnew23.html#explicit-backing-fields

topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+   backStack = backStack,
+   entryProvider = entryProvider { }
+)
+```
+
+## ViewModel and dependency injection
+
+A dependency injection (DI) framework allows you to inject different dependencies into components
+based on the current environment or target platform.
+To manage ViewModels, you can use Koin, Metro, or another DI framework that supports Kotlin Multiplatform.
```
zamulla:
```suggestion
To manage ViewModels, you can use Koin, Metro, or any other DI framework that supports Kotlin Multiplatform.
```

topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+   }
+}
+```
+
+### Shared ViewModel and platform-specific UI
+
+In this approach, the `ViewModel` (business logic) is shared, but platforms have native UI implementations.
+Learn more in [Set up ViewModel for Kotlin Multiplatform](https://developer.android.com/kotlin/multiplatform/viewmodel).
+
+1. Since the UI is not shared in this case, you can switch from the Compose Multiplatform version of the ViewModel library to the `androidx.lifecycle` library.
```
zamulla: Exactly since the UI is not shared, I would separate the Android instructions from iOS, because they can be implemented independently. So, `#### Android` and `#### iOS` sections. This also gets rid of the nested list for iOS.

topics/compose/compose-viewmodel.md
```diff
@@ -1,69 +1,465 @@
...
+Every time you create a `ViewModel` instance, you need to provide at least an initializer as an argument.
+
+If only an initializer is provided, Compose Multiplatform creates a default factory under the hood.
+However, you can implement your own factories and call more explicit versions of the common `viewModel()` function,
+just like [with Jetpack Compose](https://developer.android.com/topic/libraries/architecture/viewmodel#jetpack-compose).
+
+Let's define a ViewModel and wire it into a composable:
+
+1. Define an `OrderViewModel` class, 
+   based on the [navigation sample](https://github.com/JetBrains/compose-multiplatform/tree/0e38f58b42d23ff6d0ad30b119d34fa1cd6ccedb/examples/nav_cupcake):
```
zamulla: The sample is a bit on the older side now, and didn't let me smoothly follow the instructions after simply checking out and importing it (if that's what's implied). Ideally we update it to use the backing fields and newer dependencies so that there's less friction on the way to making it work.

zamulla (commented):
That's a big one. I left surface-level comments generally, my main problem is that there is no ready-to-go sample, and applying example code to nav_cupcake was not straightforward for me — although it did work in the end.

zamulla (approved):
👍

PR #693, opened 2026-05-14
Liquid Glass in a Compose Multiplatform app

ElviraMustafina:
TBD: the `lg-nav` branch will eventually be merged into `main`, we should agree on a plan for that


topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
+[//]: # (title: Liquid Glass in a Compose Multiplatform app)
+<show-structure depth="1"/>
+
+Liquid Glass is Apple's new visual design introduced in iOS 26, bringing translucency, depth, 
+and dynamic blur to UI elements. 
+To adopt it in a Compose Multiplatform app, you need a native SwiftUI shell, because 
+Liquid Glass effects are rendered by the system through native `TabView`, `NavigationStack`, and toolbar APIs.
```
kropp, line 7: Other controls, such as buttons, also have liquid glass effects, but this sentence reads like the effects are limited to navigation primitives.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+but the same pattern scales to any number of tabs.
+
+## Migration plan
+
+In a standard Compose Multiplatform setup, a single `ComposeUIViewController` owns the entire iOS app:
+tabs, navigation stack, back gestures, and screen content. 
+
+This approach works, but with trade-offs:
+
+* The tab bar and navigation transitions look like Material 3, not native iOS.
```
kropp: Let's rephrase it. Nav3 transitions on iOS are tuned to resemble native transitions, so it is not correct to call them material-like. Current phrasing also sheds the light on having Compose based navigation at all, that's not what we want.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+* You can't use iOS 26's Liquid Glass tab bar or system `NavigationStack` behavior.
```
kropp: Even with the native navigation approach implemented in the branch in KotlinConf app and here, some native behaviors are unsupported, such as minimizing tab bar buttons when scrolling the view.
Need to carefully avoid this in phrasing

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+* [Expose new iOS entry points](#expose-new-ios-entry-points) for creating the root view controller and individual screen view controllers.
+
+In the native iOS code (Swift):
+
+* [Build the SwiftUI navigation layer](#build-the-swiftui-navigation-layer) with `TabView`, per-tab `NavigationStack`, and the bridges that embed Compose screens.
+
+## Add title metadata to routes
+
+SwiftUI's `NavigationStack` displays a navigation bar title for each pushed destination. 
+We'll store titles directly on the route objects,
```
kropp: These titles are also needed for a back stack displayed on long press on the native Back button.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+## Alternative approaches
+
+The migration in this tutorial favors native SwiftUI navigation, 
+which gives you Liquid Glass and other system behaviors out of the box. 
+If this approach doesn't fit your project, consider one of these alternatives:
+
+* **Compose-driven navigation with native interop controls**. Keep navigation in Compose, but embed native UI controls 
+  such as `UITabBar` and `UINavigationBar` for a closer-to-native look. 
+  This avoids the SwiftUI layer entirely but doesn't give you Liquid Glass styling.
```
kropp: Why so?

kropp (changes requested):

topics/ios-liquid-glass.md
ElviraMustafina: Oops! Replaced with a mention of interop limitations between native overlays and Compose content. 
Is there anything more specific we should say about this approach?

ASalavei (approved):

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
+[//]: # (title: Liquid Glass in a Compose Multiplatform app)
+<show-structure depth="1"/>
+
+Liquid Glass is Apple's new visual design introduced in iOS 26, bringing translucency, depth,
```
EmilFlach: I would remove "new" in this sentence, it will age quickly (and is already a year old).

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+The migration in this tutorial favors native SwiftUI navigation, 
+which gives you Liquid Glass and other system behaviors out of the box. 
+If this approach doesn't fit your project, consider one of these alternatives:
+
+* **Compose-driven navigation with native interop controls**. Keep navigation in Compose, but embed native UI controls 
+  such as `UITabBar` and `UINavigationBar`, including Liquid Glass styling. The trade-off is some interop limitations 
+  between native overlays and Compose content.
+* **Compose-only navigation with imitated Liquid Glass effects**. Render everything in Compose and approximate Liquid Glass 
+  visually, for example, with a library like [KMP Liquid Glass](https://github.com/Kashif-E/KMPLiquidGlass).
```
EmilFlach:
```suggestion
  visually, for example, with libraries like [AndroidLiquidGlass](https://klibs.io/project/Kyant0/AndroidLiquidGlass),[Calf](https://klibs.io/project/MohamedRejeb/Calf) or [Liquid](https://klibs.io/project/FletchMcKee/liquid).
```

EmilFlach (approved):
Generally looks great, always love to see samples where I can follow along with a git diff.

kropp (approved):

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+iOS 26 Liquid Glass styling, while keeping Compose in charge of rendering each screen's content.
+The Liquid Glass effects themselves are applied automatically by the system once the app uses native `TabView` and `NavigationStack`,
+so you don't need to write any Liquid Glass-specific code.
+
+We'll use the official KotlinConf app as our example:
+
+![Shared UI with Material 3 design](ios-kotlinconf-no-liquid-glass.png){ width="250" style="inline"}
+![Native iOS UI with Liquid Glass](ios-kotlinconf-liquid-glass.png){ width="250" style="inline"}
+
+* [KotlinConfApp](https://github.com/JetBrains/kotlinconf-app) — sample project.
```
zamulla: We can probably skip this link since it leads to the same place the next one does.
What we can do instead actually is link the app itself in app stores above, at the first mention of KotlinConf app.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
+[//]: # (title: Liquid Glass in a Compose Multiplatform app)
+<show-structure depth="1"/>
+
+Liquid Glass is Apple's visual design introduced in iOS 26, bringing translucency, depth,
```
zamulla: Should we just throw in the [link](https://developer.apple.com/documentation/TechnologyOverviews/liquid-glass) here, too? I see it at the end, but maybe people would like to familiarize themselves with the design principles from the source.

zamulla: In the same vein, "bringing translucency and dynamic blur" doesn't sound like something entirely new for Apple — maybe we can use more specific language here. "Glass-like visuals with a new design system"? Even the developers' page is mostly about getting onboard rather than specific benefits, but still.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+<show-structure depth="1"/>
+
+Liquid Glass is Apple's visual design introduced in iOS 26, bringing translucency, depth, 
+and dynamic blur to UI elements. 
+To adopt it in a Compose Multiplatform app, you need a native SwiftUI shell, because 
+Liquid Glass effects are rendered by the system through native `TabView`, `NavigationStack`, and toolbar APIs.
+
+This tutorial walks you through migrating an iOS app from fully Compose-driven navigation to native SwiftUI navigation with
+iOS 26 Liquid Glass styling, while keeping Compose in charge of rendering each screen's content.
+The Liquid Glass effects themselves are applied automatically by the system once the app uses native `TabView` and `NavigationStack`,
```
zamulla: Active voice here? Also `TabView` and `NavigationStack` need some kind of a noun because it's not clear what they are — APIs?

```suggestion
The system applies the Liquid Glass effects automatically when the app uses native `TabView` and `NavigationStack`,
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+![Native iOS UI with Liquid Glass](ios-kotlinconf-liquid-glass.png){ width="250" style="inline"}
+
+* [KotlinConfApp](https://github.com/JetBrains/kotlinconf-app) — sample project.
+* [`main` branch](https://github.com/JetBrains/kotlinconf-app/tree/main) — starting state, with the Material 3 Expressive design.
+* [`lg-nav` branch](https://github.com/JetBrains/kotlinconf-app/tree/lg-nav) — final state, with the Liquid Glass design.
+  
+Clone the repo and check out either branch to follow along, or compare them side by side: 
+[`main...lg-nav`](https://github.com/JetBrains/kotlinconf-app/compare/main...lg-nav). You'll need Xcode 26 or later with the iOS 26 SDK.
+
+For simplicity, we'll migrate a two-tab version of the app (**Schedule** and **Info**),
```
zamulla, line 29:
```suggestion
For simplicity, we'll implement a two-tab version of the app (**Schedule** and **Info**), 
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+  
+Clone the repo and check out either branch to follow along, or compare them side by side: 
+[`main...lg-nav`](https://github.com/JetBrains/kotlinconf-app/compare/main...lg-nav). You'll need Xcode 26 or later with the iOS 26 SDK.
+
+For simplicity, we'll migrate a two-tab version of the app (**Schedule** and **Info**), 
+but the same pattern scales to any number of tabs.
+
+## Migration plan
+
+In a standard Compose Multiplatform setup, a single `ComposeUIViewController` owns the entire iOS app:
```
zamulla: "Standard" feels vague to me; should we say "in a setup with fully shared UI code"?

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+Swift: ContentView
+  └── TabView  (Liquid Glass, iOS 26)
+        ├── Tab: Schedule
+        │     └── NavigationStack
+        │           ├── NativeNavComposeView  ← Kotlin tab root
+        │           └── DetailComposeView     ← Kotlin detail screen, one per destination
+        └── Tab: Info
+              └── NavigationStack
+                    ├── NativeNavComposeView  ← Kotlin tab root
+                    └── DetailComposeView     ← Kotlin detail screen, one per destination
```
zamulla: Kotlin screen or Compose screen?

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+```
+
+Here's how navigation flows in the new setup:
+
+* SwiftUI creates a `TabView` and a `NavigationStack` per tab.
```
zamulla:
```suggestion
* SwiftUI creates a `NavigationStack` for each tab and a `TabView`.
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+    ```kotlin
+    @Serializable
+    data class SessionScreen(
+        val sessionId: SessionId,
+        override val title: String? = null,
+    ) : AppRoute
+    ```
+   
+   For the full set of updated route definitions, see [`Routes.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/Routes.kt).
```
zamulla: I think this link to be below the data object / data class explanation? So that people who follow the link understand the reason for that change.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+        </code-block>
+    </compare>
+
+4. Pass the localized title at the call site in the `NavHost.kt` file.
+   Since `stringResource` is a `@Composable` function, resolve it inside the entry scope and capture it in the click callback,
+   not inside the callback itself:
+
+    ```kotlin
+    entry<InfoScreen> {
+        val settingsTitle = stringResource(Res.string.info_link_settings)
```
zamulla:
```suggestion
        val settingsTitle = stringResource(Res.string.settings_title)
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+* Detail pushes → Swift. Whenever a non-root route lands on the back stack, 
+  forward it through `onNavigate` and remove it from Compose's back stack so SwiftUI's `NavigationStack` becomes the single source of truth.
+* Tab switches → Swift. When the top-level route changes from inside Compose, notify Swift via `onActivate` so the SwiftUI `TabView` selection stays in sync.
+
+This step is specific to the Navigation 3 library. 
+The same interception pattern applies to any Compose navigation library, 
+but the exact API (back stack access, current destination observation) will differ.
+
+In `navigation/NavHost.kt`, add the new parameters and the two interception effects:
```
zamulla:
```suggestion
In `navigation/NavHost.kt`, add the new parameters and the two interception effects to the overloaded `NavHost()` function:
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+}
+```
+
+For the full file, see [`NavHost.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/NavHost.kt).
+
+## Build a standalone screen renderer for iOS
+
+When SwiftUI owns the `NavigationStack`, Compose only needs to render the content of each pushed screen.
+`NavHost` is built for managing a back stack, transitions, and lifecycle, 
+so we need a lighter-weight entry point for rendering a single route.
```
zamulla:
```suggestion
so we need a simpler entry point for rendering a single route.
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+### Add a flat screen renderer
+
+`ScreenContent` is that smaller entry point: a flat `when` expression that maps a single detail route to its composable,
```
zamulla:
```suggestion
`ScreenContent` is that simpler entry point: a flat `when` expression that maps a single detail route to its composable,
```

Regardless of the previous suggestion, it's easier to make a callback by using exactly the same word.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+```
+
+Titles don't appear in this function: they were attached to the route objects back in the
+[Add title metadata to routes](#add-title-metadata-to-routes) step, 
+so Swift reads `wrapper.route.title` directly when configuring the navigation bar.
+
+### Signal to Compose that SwiftUI owns navigation
+
+`ScreenContent` runs in a context where SwiftUI renders the navigation bar and back button. 
+Compose screens that draw their own title bars or back buttons must skip that chrome.
```
zamulla: Not sure we introduced the term "chrome" before :)

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+To avoid duplication inside the composition tree, use a `CompositionLocal`.
+Each screen reads it locally without depending on iOS-specific code.
+
+Declare it in `NavHost.kt`, before the `NavHost` function:
```
zamulla:
```suggestion
Declare it in the `NavHost.kt` file, before the `NavHost()` function:
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+### Wrap the renderer for iOS
+
+`ScreenContent` renders a route, but it needs a wrapper that sets the same theme, dependency injection,
+and app-wide composition locals that `App()` usually sets up.
+
+Add the `SingleScreenApp` wrapper. It mirrors the setup from `App()` and additionally sets 
+`LocalUseNativeNavigation` to `true`, so each screen automatically hides its Compose-rendered title bar 
+and back button (introduced in the [Hide Compose's built-in navigation UI](#hide-compose-s-built-in-navigation-ui) step).
+
+In the `iosMain` source set, create `SingleScreenApp.kt`:
```
zamulla:
```suggestion
In the `iosMain` source set, create the `SingleScreenApp.kt` file: 
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+}
+```
+
+For the full implementation, see [`App.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/App.kt).
+
+## Intercept navigation at the Compose level
+
+Now that `App()` exposes navigation callbacks, `NavHost` needs to use them. 
+Whenever a detail route appears on Compose's back stack, hand it off to Swift and immediately remove 
+it from Compose. This way, Compose only renders detail screens only when invoked from Swift.
```
zamulla:
```suggestion
it from Compose. This way, Compose renders detail screens only when invoked from Swift.
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+                ScreenContent(route, onNavigate, onGoBack, onSet, onActivate)
+            }
+        }
+    }
+}
+```
+
+### Apply the flag to tab roots
+
+Tab roots still go through the regular `NavHost`, so they also need to honor `LocalUseNativeNavigation`.
```
zamulla:
```suggestion
Tab roots still go through the regular `NavHost`, so they also need to honor the `LocalUseNativeNavigation` value.
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+```
+
+For full implementations, see [`NavHost.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/NavHost.kt)
+and [`SingleScreenApp.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/iosMain/kotlin/org/jetbrains/kotlinconf/SingleScreenApp.kt).
+
+## Hide Compose's built-in navigation UI
+
+With `LocalUseNativeNavigation` set wherever SwiftUI owns navigation, 
+individual screens now need to read it and hide their own chrome. 
+Otherwise, the user sees two title bars stacked on top of each other and two competing back buttons.
```
zamulla:
```suggestion
Otherwise, the user would see two title bars stacked on top of each other and two competing back buttons.
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+Apply the same pattern to any other screens that draw their own back buttons or headers.
+
+For the full implementation, see [`BaseScreens.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/BaseScreens.kt).
+
+## Expose new iOS entry points
+
+Swift needs three Kotlin entry points to build the new navigation structure:
+
+* `MainViewController` with callbacks, for each tab root.
```
zamulla:
```suggestion
* `MainViewController` for each tab root, with corresponding callbacks.
```

Not sure I understand this item correctly, which is the main problem, but I guessed the fix :)

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+For full implementations, see [`NavHost.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/navigation/NavHost.kt)
+and [`SingleScreenApp.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/iosMain/kotlin/org/jetbrains/kotlinconf/SingleScreenApp.kt).
+
+## Hide Compose's built-in navigation UI
+
+With `LocalUseNativeNavigation` set wherever SwiftUI owns navigation, 
+individual screens now need to read it and hide their own chrome. 
+Otherwise, the user sees two title bars stacked on top of each other and two competing back buttons.
+
+In `BaseScreens.kt`, wrap the title bar and divider in `ScreenWithTitle` so they're skipped when Swift is in charge:
```
zamulla:
```suggestion
In `BaseScreens.kt`, wrap the title bar and divider in the body of the `ScreenWithTitle()` function so they're skipped when Swift is in charge:
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+For the full implementation, see [`BaseScreens.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/BaseScreens.kt).
+
+## Expose new iOS entry points
+
+Swift needs three Kotlin entry points to build the new navigation structure:
+
+* `MainViewController` with callbacks, for each tab root. 
+  Compose runs the full `App()` and `NavHost`, but navigation events are forwarded to Swift instead of being handled internally.
+* `ScreenViewController`, for each pushed detail screen. Renders a single route via `SingleScreenApp`, with `LocalUseNativeNavigation = true` so Compose chrome is hidden.
+* `MainViewController` without callbacks, as a pre-iOS 26 fallback. Liquid Glass APIs require iOS 26, so Swift falls back
```
zamulla:
```suggestion
* `MainViewController` without callbacks, as a pre-iOS 26 fallback. Liquid Glass APIs require iOS 26, so Swift should fall back 
```

If I understand correctly, that's why we need to implement a fallback, it won't do it by itself.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+fun MainViewController(
+    topLevelRoute: TopLevelRoute,
+    onNavigate: (AppRoute) -> Unit,
+    onGoBack: () -> Unit,
+    onSet: (AppRoute) -> Unit,
+    onActivate: (TopLevelRoute) -> Unit,
+): UIViewController = ComposeUIViewController(
+    configure = { onFocusBehavior = OnFocusBehavior.DoNothing }
+) {
+    // `onGoBack` and `onSet` were for API symmetry with `ScreenViewController`
```
zamulla: I think something is missing after "were".

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+This is the iOS side of the migration. All the Kotlin changes from the previous steps prepare the app for what happens here: 
+a SwiftUI `TabView` with per-tab `NavigationStack`s that host Compose views as their destinations.
+To build that, complete the following:
+
+1. [Make Kotlin routes usable in `NavigationStack`](#make-kotlin-routes-usable-in-navigationstack).
+2. [Track tab and navigation state](#track-tab-and-navigation-state).
+3. [Embed Compose screens as SwiftUI views](#embed-compose-screens-as-swiftui-views).
+4. [Set up navigation within each tab](#set-up-navigation-within-each-tab).
+5. [Build the tab bar](#build-the-tab-bar).
+6. [Fall back on older iOS versions](#fall-back-on-older-ios-versions).
```
zamulla: The guidelines guide us away from periods at the end of list items like these.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+The Liquid Glass effects themselves are applied automatically by the system once the app uses native `TabView` and `NavigationStack`,
+so you don't need to write any Liquid Glass-specific code.
+
+We'll use the official KotlinConf app as our example:
+
+![Shared UI with Material 3 design](ios-kotlinconf-no-liquid-glass.png){ width="250" style="inline"}
+![Native iOS UI with Liquid Glass](ios-kotlinconf-liquid-glass.png){ width="250" style="inline"}
+
+* [KotlinConfApp](https://github.com/JetBrains/kotlinconf-app) — sample project.
+* [`main` branch](https://github.com/JetBrains/kotlinconf-app/tree/main) — starting state, with the Material 3 Expressive design.
```
kropp: KotlinConf app doesn't use Material at all, it would be more correct to say "starting state, with a custom theme fully implemented in Compose" or something like that

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+1. [Make Kotlin routes usable in `NavigationStack`](#make-kotlin-routes-usable-in-navigationstack).
+2. [Track tab and navigation state](#track-tab-and-navigation-state).
+3. [Embed Compose screens as SwiftUI views](#embed-compose-screens-as-swiftui-views).
+4. [Set up navigation within each tab](#set-up-navigation-within-each-tab).
+5. [Build the tab bar](#build-the-tab-bar).
+6. [Fall back on older iOS versions](#fall-back-on-older-ios-versions).
+
+Note that none of the code in this section applies Liquid Glass effects directly. 
+iOS 26 renders Liquid Glass automatically for native `TabView` and `NavigationStack`,
+and the migration itself is what unlocks it.
```
zamulla: I myself was a little confused here because it's a long way down in the article and I'm not sure anymore which migration we're talking about. Maybe simply "...and using them is what unlocks it"?

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+### Make Kotlin routes usable in `NavigationStack`
+
+`NavigationStack` requires its path elements to be `Hashable` and identifiable.
```
zamulla:
```suggestion
`NavigationStack` requires its path elements to be `Hashable` and `Identifiable`.
```

Why not both? :)

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+`AppRoute` is a Kotlin sealed interface, so wrap it in a Swift struct:
```
zamulla:
```suggestion
`AppRoute` is a Kotlin sealed interface, so wrap it in a Swift structure:
```

Either that, or wrap it in backticks, too.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(id)
+    }
+}
+```
+
+Pushing the same route twice should create two distinct stack entries, 
+matching the expected navigation behavior. 
+To achieve this, identity is based on `UUID` rather than the route's content.
```
zamulla:
```suggestion
To achieve this, identity is based on a UUID rather than the route's content.
```
Here it's the other way, UUID is a language-agnostic concept.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+    func activateTab(for route: TopLevelRoute) {
+        if route is ScheduleScreen {
+            selectedTab = .schedule
+        } else if route is InfoScreen {
+            selectedTab = .info
+        }
+    }
+}
+```
+{initial-collapse-state="collapsed" collapsible="true" collapsed-title="@Observable class AppNavigationCoordinator { "}
```
zamulla, line 546: Should we comment on the fact that this implementation is different from the actual branch?

zamulla: I guess I noticed that we stopped linking specific files with implementations and not sure when (if) this is about the fact that here we only implement two tabs while the branch implements everything.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+                }
+        }
+    }
+}
+```
+
+### Build the tab bar
+
+The top-level container is a `TabView` with one `Tab` per top-level route.
+`.tabBarMinimizeBehavior(.automatic)` enables the Liquid Glass floating tab bar.
```
zamulla:
```suggestion
The `.tabBarMinimizeBehavior(.automatic)` call makes the tab bar float and minimize on scroll. Without it, the tab bar stays fixed at the bottom. Both variants render with Liquid Glass on iOS 26. 
```

Right? Since there isn't a "Liquid Glass tab bar" and on iOS 26 the native tab bars are styled automatically.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+            onActivate: { route in self.appCoordinator.activateTab(for: route) }
+        )
+    }
+
+    func updateUIViewController(_ uiViewController: UIViewController, context: Context) {
+    }
+}
+```
+
+`DetailComposeView` hosts a single detail screen, one instance per `NavigationStack` destination:
```
zamulla:
```suggestion
`DetailComposeView` hosts a single detail screen, one instance for each `NavigationStack` destination:
```

Here and elsewhere, "per" is usually used to describe rates in measurement units — it's not confusing, but could be clearer I think.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+                               appCoordinator: appCoordinator, title: String(localized: "Info"))
+            }
+        }
+        .tabBarMinimizeBehavior(.automatic) 
+        .tint(Color(.accent))
+    }
+}
+```
+
+`.tint(Color(.accent))` applies an accent color to the tab bar.
```
zamulla:
```suggestion
The `.tint(Color(.accent))` call applies an accent color to the tab bar. 
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+        .tabBarMinimizeBehavior(.automatic) 
+        .tint(Color(.accent))
+    }
+}
+```
+
+`.tint(Color(.accent))` applies an accent color to the tab bar. 
+`Color(.accent)` resolves to the `AccentColor` asset in your Xcode project's asset catalog. 
+Define this asset by creating `Assets.xcassets/AccentColor.colorset/Contents.json`: you can use [`Contents.json`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/iosApp/iosApp/Assets.xcassets/AccentColor.colorset/Contents.json)
+from the sample project as a starting point and replace the component values with your own colors.
```
zamulla, line 704: I think we can also link the Xcode UI for this: https://developer.apple.com/documentation/xcode/specifying-your-apps-color-scheme

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+The migration in this tutorial favors native SwiftUI navigation, 
+which gives you Liquid Glass and other system behaviors out of the box. 
+If this approach doesn't fit your project, consider one of these alternatives:
+
+* **Compose-driven navigation with native interop controls**. Keep navigation in Compose, but embed native UI controls 
+  such as `UITabBar` and `UINavigationBar`, including Liquid Glass styling. The trade-off is some interop limitations 
+  between native overlays and Compose content.
+* **Compose-only navigation with imitated Liquid Glass effects**. Render everything in Compose and approximate Liquid Glass 
+  visually, for example, with libraries like [AndroidLiquidGlass](https://klibs.io/project/Kyant0/AndroidLiquidGlass),[Calf](https://klibs.io/project/MohamedRejeb/Calf) or [Liquid](https://klibs.io/project/FletchMcKee/liquid).
+  This approach keeps all UI on the Compose side, with the effect visually similar but not identical to system Liquid Glass.
```
zamulla:
```suggestion
  This approach keeps all UI on the Compose side, with the effect visually similar although not identical to system Liquid Glass.
```
Juuust a tad softer.

topics/ios-liquid-glass.md
zamulla: Also we should refer, as in other instructions, to the specific file (ContenView.swift?)

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+}
+```
+
+Apply the same pattern to any other screens that draw their own back buttons or headers.
+
+For the full implementation, see [`BaseScreens.kt`](https://github.com/JetBrains/kotlinconf-app/blob/3982334f1c3712fb959f0d20b563d6c8b81e9bbd/app/shared/src/commonMain/kotlin/org/jetbrains/kotlinconf/BaseScreens.kt).
+
+## Expose new iOS entry points
+
+Swift needs three Kotlin entry points to build the new navigation structure:
```
zamulla: Should the order in the list match the order in the code below? It can be a little disorienting to match them.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+        )
+        // All other detail routes
+        else -> {}
+    }
+}
+```
+
+Titles don't appear in this function: they were attached to the route objects back in the
+[Add title metadata to routes](#add-title-metadata-to-routes) step, 
+so Swift reads `wrapper.route.title` directly when configuring the navigation bar.
```
zamulla: This refers to Swift code that is introduced way later. Should we talk in the abstract here?

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+On older versions, the app falls back to the previous Compose-driven setup using the no-callback `MainViewController`
+overload added in the previous step:
+
+```swift
+struct ContentView: View {
+    var body: some View {
+        if #available(iOS 26.0, *) {
+            NativeNavContentView()
+        } else {
+            ComposeView(topLevelRoute: ScheduleScreen())
```
zamulla, line 724: Since we don't mention `ComposeView()`, maybe we should leave a comment here explaining what it is? It's there in the full file, but just to wrap it up neatly.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+            selectedTab = .info
+        }
+    }
+}
+```
+{initial-collapse-state="collapsed" collapsible="true" collapsed-title="@Observable class AppNavigationCoordinator { "}
+
+### Embed Compose screens as SwiftUI views
+
+Two `UIViewControllerRepresentable` types connect the Kotlin entry points from the previous step to SwiftUI:
```
zamulla: Maybe a link for "previous step"? There are a lot of steps :)

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+* When the user triggers navigation from a Compose screen (for example, taps a list row), the event is forwarded to Swift via `onNavigate`.
+* The Swift coordinator pushes the route onto its `NavigationStack`, which creates a new `UIViewController` hosting a single Compose screen.
+
+The migration touches both the shared Compose Multiplatform code and the native iOS code.
+In the shared Kotlin code:
+
+* [Add title metadata to routes](#add-title-metadata-to-routes) so SwiftUI can render navigation bar titles and back stack entries without calling back into Kotlin.
+* [Add navigation callbacks to the iOS entry point](#add-navigation-callbacks-to-the-ios-entry-point) so the iOS layer can control which tab is active and respond to navigation events.
+* [Intercept navigation at the Compose level](#intercept-navigation-at-the-compose-level) so detail routes are forwarded to Swift instead of being handled by Compose.
+  This tutorial shows the Nav3 implementation — adapt this step if you use a different navigation library.
```
zamulla:
```suggestion
  This tutorial shows the Navigation 3 implementation — adapt this step if you use a different navigation library.
```

If we're not going to use the shortened version more than once, let's not use it :)

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+```
+
+### Wrap the renderer for iOS
+
+`ScreenContent` renders a route, but it needs a wrapper that sets the same theme, dependency injection,
+and app-wide composition locals that `App()` usually sets up.
+
+Add the `SingleScreenApp` wrapper. It mirrors the setup from `App()` and additionally sets 
+`LocalUseNativeNavigation` to `true`, so each screen automatically hides its Compose-rendered title bar 
+and back button (introduced in the [Hide Compose's built-in navigation UI](#hide-compose-s-built-in-navigation-ui) step).
```
zamulla: A reference to the next step in the past tense, breaks the flow a bit.

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,734 @@
...
+
+![Shared UI with Material 3 design](ios-kotlinconf-no-liquid-glass.png){ width="250" style="inline"}
+![Native iOS UI with Liquid Glass](ios-kotlinconf-liquid-glass.png){ width="250" style="inline"}
+
+* [KotlinConfApp](https://github.com/JetBrains/kotlinconf-app) — sample project.
+* [`main` branch](https://github.com/JetBrains/kotlinconf-app/tree/main) — starting state, with the Material 3 Expressive design.
+* [`lg-nav` branch](https://github.com/JetBrains/kotlinconf-app/tree/lg-nav) — final state, with the Liquid Glass design.
+  
+Clone the repo and check out either branch to follow along, or compare them side by side: 
+[`main...lg-nav`](https://github.com/JetBrains/kotlinconf-app/compare/main...lg-nav). You'll need Xcode 26 or later with the iOS 26 SDK.
```
zamulla: We should probably set this requirement as a note before going into specific repository.

zamulla (commented):
Huge work!

Some comments™

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,735 @@
...
+                    ├── NativeNavComposeView  ← Kotlin tab root
+                    └── DetailComposeView     ← Kotlin detail screen, one per destination
+```
+
+Here's how navigation flows in the new setup:
+
+* SwiftUI creates a `TabView` and a `NavigationStack` per tab.
+* Compose still renders each screen's content but no longer manages the back stack.
+* When the user triggers navigation from a Compose screen (for example, taps a list row), the event is forwarded to Swift via `onNavigate`.
+* The Swift coordinator pushes the route onto its `NavigationStack`, which creates a new `UIViewController` hosting a single Compose screen.
```
zamulla:
```suggestion
* The Swift coordinator pushes the route into its `NavigationStack`, which creates a new `UIViewController` hosting a single Compose screen.
```

topics/ios-liquid-glass.md
```diff
@@ -0,0 +1,753 @@
...
+    fun MainViewController(topLevelRoute: TopLevelRoute): UIViewController = ComposeUIViewController(
+        configure = { onFocusBehavior = OnFocusBehavior.DoNothing },
+    ) {
+        App(appGraph, topLevelRoute)
+    }
+    ```
+
+* `MainViewController` with callbacks, called by SwiftUI for each tab root. 
+  Compose runs the full `App()` and `NavHost`, but navigation events are forwarded to SwiftUI instead of being handled internally.
+  The signature includes `onGoBack` and `onSet` for API symmetry with `ScreenViewController`, though they aren't used in this overload.
```
zamulla:
```suggestion
  The signature includes `onGoBack` and `onSet` for API symmetry with `ScreenViewController`, although they aren't used in this overload.
```

zamulla (approved):
All good!

PR #699, opened 2026-05-18
 Correct desktop app to web app in documentation

I-am-Abhilash:
The configuration steps are for the web target, but the descriptions trailing the code blocks still had remnants of the desktop guide.


zamulla (approved):
Thanks for catching that!

PR #701, opened 2026-05-19
feat: Stabilize Uuid API

aepanchi:
Update mentions of the UUID class to address its stabilization. 
Should NOT be merged until https://github.com/JetBrains/kotlin-web-site/pull/5537 is published.


zamulla (approved):
👍

PR #714, opened 2026-06-04
Upd: gifs and screenshots for Compose Hot Reload

ElviraMustafina:


topics/tools/compose-hot-reload.md
```diff
@@ -101,7 +101,7 @@ The steps refer to the project from the [Create an app with shared logic and UI]
 
 ## Use Compose Hot Reload
 
-1. In the `jvmMain` directory, open the `main.kt` file and update the `main()` function:
+1. In the `desktopApp` directory, open the `main.kt` file and update the `main()` function:
```
zamulla: The file is not in the `desktopApp` directory, but it's somewhere in there, and I don't think we should write out the path. Maybe something like "Within the `desktopApp` directory"? Not sure if that's better :)

zamulla (approved):
One tiniest comment.

PR #719, opened 2026-06-12
fix: correct missing ref and references

louisunlimited:
When I was going through the flow I found a couple mismatched refs and typos that prevented me from building the app. Not sure if this is the right fix for the doc but doing it this way helped me progress through the tutorial here: https://kotlinlang.org/docs/multiplatform/multiplatform-create-first-app.html


zamulla (approved):
Thanks for catching this one!

PR #732, opened 2026-07-07
fix: split  toc-element into visible link and hidden redirect

berezinant:
After updating the version of writerside builder the check TOC020 (Target for external redirect must be a hidden TOC element) became strict (example of [failure](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinSites_KotlinlangTeamcityDsl_KotlinMultiplatform/996164469?buildTab=log&logView=flowAware&linesState=1293.2408&focusLine=2429) ), so we need this fix in order to deploy docs with new builder


nikpachoo (approved):

zamulla (approved):
👍

PR #734, opened 2026-07-07
upd: set ProGuard version

ElviraMustafina:


topics/compose/compose-native-distribution.md
```diff
@@ -925,65 +925,85 @@ For each *default* packaging task (without ProGuard), the Gradle plugin provides
...
-
-```kotlin
-compose.desktop {
-    application {
-        buildTypes.release.proguard {
-            joinOutputJars.set(true)
+    ```
+    {initial-collapse-state="collapsed" collapsible="true" collapsed-title="optimize.set(false)"}
+
+* `joinOutputJars` produces a single uber JAR. By default,
```
zamulla:
```suggestion
* `joinOutputJars` produces a single uber-JAR. By default, 
```

topics/compose/compose-native-distribution.md
```diff
@@ -925,65 +925,85 @@ For each *default* packaging task (without ProGuard), the Gradle plugin provides
...
     }
-}
-```
+    ```
+    {initial-collapse-state="collapsed" collapsible="true" collapsed-title="joinOutputJars.set(true)"}
+
+[//]: # (TODO update version for stable release)
+
+* `version` sets a specific ProGuard version. Starting with Compose Multiplatform 1.12.0-beta01, the default ProGuard version is 7.8.0.
+  If you use an earlier version of Compose Multiplatform and build with JDK 25, explicitly set the version to `7.8.0`:
```
zamulla: What if I don't? Is it JDK 25 that requires 7.8.0, should we say that explicitly?

zamulla (approved):
👍 
A couple of superficial comments.

PR #740, opened 2026-07-18
Fix uiState reference in snippet

vladcudoidem:
There was a small typo in the code snippet.

No idea why the diff says that I changed line 473 as well; the contents there are the same.


ElviraMustafina (approved):
Oops! Thanks for catching this 👍

PR #749, opened 2026-07-23
update: updating two images

Daria-Voronina:
After internal feedback, we agreed to update the two images. Please take a look.


zamulla (approved):
👍

PR #753, opened 2026-07-30
Retroactively Update What's New pages re. skiko.js requirements

SebastianAigner:
Since this is something that cost me a while during our own updates:

As of Compose Multiplatform 1.9.0, skiko.js is no longer required for Kotlin/JS either. It was already redundant for Kotlin/Wasm since 1.7.0. Added a short section to the 1.9.3 what's new page for this, and pointed the old note on the 1.7.0 page at it.

This info was previously somewhat hard to find, only in the changelog, which doesn't readily appear in Google searches.


zamulla (approved):
Thanks!

PR #755, opened 2026-08-05
update: small update in KMP for iOS in Journal 

Daria-Voronina:
Adding "realities" right after myths, fixing levels of two subheadings.


zamulla (approved):
👍

PR #763, opened 2026-08-20
Languages with multiple scripts

ElviraMustafina:


terrakok (approved):

topics/compose/compose-localize-strings.md
```diff
@@ -27,6 +27,11 @@ commonMain/composeResources/
 └── ... (other locale directories)
 ```
 
+If a language has more than one script, such as Chinese or Serbian, add a script-less
```
zamulla:
```suggestion
If a language has more than one script (for example, Chinese or Serbian), add a script-less
```

Trying to avoid it sounding like it's the script that is Chinese rather than the language. Could also be like "Some languages, such as Chinese or Serbian, have more than one script. <End sentence so that the thought settles in.> To display them properly..."

topics/compose/compose-multiplatform-resources-setup.md
```diff
@@ -149,8 +149,31 @@ You can combine language and region qualifiers:
     The regional code must have a lowercase `r` prefix, for example: `drawable-spa-rMX`.
 
 The language and regional codes are case-sensitive.
+
+To specify a script, or any other BCP 47 subtag, use the `b+` prefix with `+` as a separator,
```
zamulla: Maybe `b` is the prefix and `+` is the separator? I had to do a double take and make sure I understand the different functions of the pluses even though they are not different really.

topics/compose/compose-multiplatform-resources-setup.md
```diff
@@ -149,8 +149,31 @@ You can combine language and region qualifiers:
...
+for example: `values-b+zh+Hans` and `values-b+zh+Hant`.
+
 Learn more about working with region-specific formats in [Localization](compose-regional-format.md).
 
+#### Fallback for languages with multiple scripts
+
+On Android and desktop, the current locale may not include script info, so the system requests
```
zamulla: Can we just say "On Android and desktop the system may request locale with an empty script, for example..."? I'm not sure what it means for a locale to include script info, and I'm not sure it matters.

topics/compose/compose-multiplatform-resources-setup.md
```diff
@@ -149,8 +149,31 @@ You can combine language and region qualifiers:
...
+`zh` with an empty script instead of specific `zh-Hans` or `zh-Hant`.
+If you provide only script-specific resources, both of the options match such a request
```
zamulla:
```suggestion
If you provide only script-specific resources, all of them match a script-less request
```

topics/compose/compose-multiplatform-resources-setup.md
```diff
@@ -149,8 +149,31 @@ You can combine language and region qualifiers:
...
+for example: `values-b+zh+Hans` and `values-b+zh+Hant`.
+
 Learn more about working with region-specific formats in [Localization](compose-regional-format.md).
 
+#### Fallback for languages with multiple scripts
+
+On Android and desktop, the current locale may not include script info, so the system requests
+`zh` with an empty script instead of specific `zh-Hans` or `zh-Hant`.
+If you provide only script-specific resources, both of the options match such a request 
+and the resource cannot be resolved, so the app fails with an error.
```
zamulla: "Fails with error" is a bit vague — is it a crash or something more graceful?

zamulla (approved):
All good, my comments hopefully would help polish it to a shine :)

topics/compose/compose-multiplatform-resources-setup.md
ElviraMustafina: Nope, it is the `b+` prefix, and then all subsequent subtag hyphens (-) must be replaced with a plus sign (+). Added details

PR #770, opened 2026-09-03
fix: replace internal API with isSystemInDarkTheme()

ElviraMustafina:


topics/compose/compose-resource-environment.md
```diff
@@ -219,7 +219,7 @@ you can address this difference using the `expect-actual` mechanism to manage pl
             } else {
                 Configuration(LocalConfiguration.current).apply {
                     uiMode = when (value) {
-                        true -> (uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_YES
+                        true  -> (uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_YES
```
zamulla:
```suggestion
                        true -> (uiMode and UI_MODE_NIGHT_MASK.inv()) or UI_MODE_NIGHT_YES
```

topics/compose/compose-resource-environment.md
```diff
@@ -229,24 +229,18 @@ you can address this difference using the `expect-actual` mechanism to manage pl
     }
     ```
 
-3. On iOS, desktop, and web platforms, you can change `LocalSystemTheme` directly:
+3. On iOS, desktop, and web platforms, add the actual implementation that overrides the theme via a `LocalIsDarkTheme` composition local with an `isSystemInDarkTheme()` fallback:
 
     ```kotlin
-    @OptIn(InternalComposeUiApi::class)
+    private val LocalIsDarkTheme = staticCompositionLocalOf<Boolean?> { null }
```
MatkovIvan: Declaring composition local is much easier
```kotlin
 val LocalIsDarkTheme = compositionLocalWithComputedDefaultOf { isSystemInDarkTheme() }
```

No expect/actual, no "object".

But the idea of this doc is to hack the system to override value that returned by `isSystemInDarkTheme` because libraries that user does not control use `isSystemInDarkTheme` directly

MatkovIvan (changes requested):

topics/compose/compose-resource-environment.md
ElviraMustafina: Is the original version acceptable for now? I reverted the changes and will update the docs once a proper public API is available

MatkovIvan: Only after https://github.com/jetbrains/compose-multiplatform-core/pull/3387 (1.12.1)

PR #773, opened 2026-09-03
fix(): point to kotlin release instead of compose in header

berezinant:
Fixes this 
Screenshot 2026-09-02 at 12-16-04
Screenshot 2026-09-02 at 12-17-13


zamulla (approved):
🙌

PR #774, opened 2026-09-07
clarify hotMcpServer name

ElviraMustafina:


topics/tools/compose-hot-reload.md
```diff
@@ -185,6 +185,13 @@ For example, in `.mcp.json`:
 }
 ```
 
+Gradle searches for tasks across all subprojects and matches the short name `hotMcpServer` to target-specific variants
+such as `hotMcpServerJvm` or `hotMcpServerDesktop`.
```
zamulla, line 189: It might not be obvious what we're talking about here, how about we point to the exact line? Something like "The last argument (`hotMcpServer`) is the name of the Gradle task. Gradle searches..."

zamulla (approved):
👍 

One suggestion to clarify.

topics/tools/compose-hot-reload.md
ElviraMustafina: It will be a repetition of the intro sentence, so I hope the current phrasing is clear enough

PR #777, opened 2026-09-17
Suggestion for the title and into of the KTC starter

amakeev:


topics/tools/kotlin-toolchain.md
```diff
@@ -1,26 +1,17 @@
-[//]: # (title: Creating and building a Kotlin Multiplatform project with Kotlin Toolchain)
+[//]: # (title: Creating and building a Kotlin Multiplatform application with Kotlin Toolchain)
 
-[Kotlin Toolchain](https://kotlin-toolchain.org/) is a tool created by JetBrains to help you configure projects
-for building, packaging, publishing, and more. With Kotlin Toolchain, you can spend less time dealing with build systems and focus
-on addressing real business challenges instead.
+[Kotlin Toolchain](https://kotlin-toolchain.org/) supports the full project lifecycle out of the box, from creation to publication, so you can focus on real business challenges. It is optimized for the agentic and CLI workflows and helps AI agents to interact with the build system reliably. It also provides traditional IDE support with plugins for IntelliJ IDEA and Android Studio.
```
zamulla:
```suggestion
[Kotlin Toolchain](https://kotlin-toolchain.org/) supports the full project lifecycle out of the box, from creation to publication, so you can focus on real business challenges. It is optimized for the agentic and CLI workflows, which helps AI agents to interact with the toolchain reliably. Kotlin Toolchain also provides traditional IDE support with plugins for IntelliJ IDEA and Android Studio.
```

I do like that "Kotlin Toolchain" does not need an explicit definition, it is what it is (e.g. not a "tool"); but we also shouldn't refer to it as a "build system" for the same reason, WDYT?

zamulla (approved):

topics/tools/kotlin-toolchain.md
amakeev: agree! I don't like build system either.

